PR #40, opened 2023-03-31
Respect namespace of current KubeConfig context

### LINKED ISSUE #39: timoni `--namespace` doesn't respect the default ns set in kubeconfig context
Hey,

I have an issue with the `timoni list` command. It should show me all the installed instances in the current namespace but the list is empty except when I explicitly use the `-A` or `-n test` flag.

**Steps to reproduce**
```bash
kubectl create namespace test
kubectl config set-context --current --namespace=test
timoni -n test apply podinfo oci://ghcr.io/stefanprodan/modules/podinfo --version latest
timoni list
```

**Expected output**
```bash
NAME   	MODULE                                    	VERSION	LAST APPLIED         
podinfo	oci://ghcr.io/stefanprodan/modules/podinfo	6.3.4  	2023-03-31T13:37:18Z
```

**Actual output**
```bash
NAME	MODULE	VERSION	LAST APPLIED
```

**Additional info**
Using the `-n test` flag explicitly or the `-A` flag works but I thought it should also work without it when I'm on the right namespace already. Is that the intended behavior or a bug?

hiddeco:
This changes the hardcoded "default" to respect the namespace of the active KubeConfig context, if usable.

Fixes #39

### DIFF
--- a/cmd/timoni/main.go
+++ b/cmd/timoni/main.go
@@ -17,14 +17,15 @@ limitations under the License.
 package main
 
 import (
-	"github.com/fluxcd/pkg/oci"
-	apiv1 "github.com/stefanprodan/timoni/api/v1alpha1"
 	"os"
 	"time"
 
+	"github.com/fluxcd/pkg/oci"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM 

Thanks @hiddeco 🥇

PR #46, opened 2023-04-02
Remove deprecated `rand.Seed`

hiddeco:
> The math/rand package now automatically seeds the global random
> number generator (used by top-level functions like Float64 and Int)
> with a random value, and the top-level Seed function has been
> deprecated.

From: https://go.dev/doc/go1.20#minor_library_changes (`math/rand`)


stefanprodan (approved):
LGTM 

Thanks @hiddeco

PR #48, opened 2023-04-02
bundle: Allow modules to be referenced by digest

stefanprodan:
This PR extends the Bundle API with support for OCI digests, the full specification can be found in #59.  

Example:

```cue
bundle: {
	apiVersion: "v1alpha1"
	instances: {
		redis: {
			module: {
				url:     "oci://<registry>/<org>/modules/redis"
				digest:  "sha256:<hex>"
			}
			namespace: "default"
		}
	}
}
```

Closes: #59

### DIFF
--- a/api/v1alpha1/bundle.go
+++ b/api/v1alpha1/bundle.go
@@ -29,6 +29,9 @@ const (
 	// BundleModuleVersionSelector is the CUE path for the Timoni's bundle module version.
 	BundleModuleVersionSelector Selector = "module.version"
 
+	// BundleModuleDigestSelector is the CUE path for the Timoni's bundle module digest.
+	BundleModuleDigestSelector Selector = "module.digest"
+
 	// BundleNamespaceSelector is the CUE path for the Timoni's bundle instance namespace.
 	BundleNamespaceSelector Selector = "namespace"
 
@@ -46,7 +49,8 @@ import "strings"
 	instances: [string & =~"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$" & strings.MaxRunes(63) & strings.MinRunes(1)]: {
 		module: close({
 			url:     string & =~"^oci://.*$"
-			version: string & strings.MinRunes(3)
+			version: *"latest" | string
+			digest?:  string
 		})
 		namespace: string & =~"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$" & strings.MaxRunes(63) & strings.MinRunes(1)
 		values: {...}
--- a/cmd/timoni/bundle_apply.go
+++ b/cmd/timoni/bundle_apply.go
@@ -27,9 +27,7 @@ import (
 	"strings"
 	"time"
 
-	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/cuecontext"
-	"cuelang.org/go/cue/load"
 	"github.com/fluxcd/pkg/ssa"
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -102,15 +100,7 @@ func runBundleApplyCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	ctx := cuecontext.New()
-
-	cfg := &load.Config{
-		Package:   "_",
-		DataFiles: true,
-	}
-
 	files := append(bundleApplyArgs.files, bundleSchema.Name())
-
 	for i, file := range files {
 		if file == "-" {
 			path, err := saveReaderToFile(cmd.InOrStdin())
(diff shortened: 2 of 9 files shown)


hiddeco (commented):
Think it would be neat if supplying both confirmed the version matches the expected digest, instead of it being an OR.

This would also kind of prevent the annoyance I have at present with GitHub Actions which also has an OR, and thereby prevents having a human readable version identifier without putting in a comment.

stefanprodan:
Uhhh I expect for Dependabot to be able to update bundles in the future, that's why I made it an OR. If the version must match the digest then it would be super hard to make Dependabot work with it I think.

hiddeco:
Including a digest and version wouldn't prevent someone from just including a digest.

stefanprodan:
With this PR you can include both but the version is ignored. The version defaults to latest if not specified which would break the execution if latest must match the digest.

hiddeco:
Sharing here in public, after having a chat in private. For the sake of having it recorded when we might run into a "why is this the way it is".

> The result of ignoring it is that you get a false sense of what’s being installed if you forget to update the version together with the SHA.
>
> I’ve had this happen with GitHub Action version comments.
>
> Mismatch between SHA and comment, thinking I installed A while it was actually B.
>
> This is nonapparent until you manually go and check (or in this example: when Dependabot updates your mistake).

cmd/timoni/bundle_apply.go
```diff
@@ -169,9 +169,22 @@ func runBundleApplyCmd(cmd *cobra.Command, args []string) error {
...
+				version = "@" + v
+			}
+		}
+
+		if version == "" {
+			moduleVersion := expr.LookupPath(cue.ParsePath(apiv1.BundleModuleVersionSelector.String()))
+			if moduleVersion.Err() != nil {
+				return fmt.Errorf("lookup %s failed, error: %w", apiv1.BundleModuleVersionSelector.String(), instances.Err())
+			}
+			if v, err := moduleVersion.String(); err == nil && v != "" {
```
squaremo, line 185: What if `err != nil` here, what does that entail?

cmd/timoni/bundle_lint.go
```diff
@@ -96,16 +96,6 @@ func runBundleLintCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	apiVersion := v.LookupPath(cue.ParsePath(apiv1.BundleAPIVersionSelector.String()))
```
squaremo, line 99: Is this removal an incidental change -- maybe it could go in its own commit, just to keep things tidy.

stefanprodan: The apiVersion is now being validated by the schema `apiVersion: string & =~"^v1alpha1$"` so there is no need to look it up in Go. If the apiVersion is not specified or if its value is not v1alpha1 linting will fail.

stefanprodan:
@hiddeco @squaremo thanks for the feedback. I have refactored the bundle validation and apply logic. Please take a look if you can.

api/v1alpha1/bundle.go
```diff
@@ -46,7 +49,8 @@ import "strings"
 	instances: [string & =~"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$" & strings.MaxRunes(63) & strings.MinRunes(1)]: {
 		module: close({
 			url:     string & =~"^oci://.*$"
-			version: string & strings.MinRunes(3)
+			version: *"latest" | string
```
squaremo, line 52: I'm not sure how defaults are applied in Cue; I suspect that if you run a value against this schema, then you get the unified version out the other end, and that will have the default applied. That means there's no distinguishing between these cases:

| version | digest | meaning |
|-----------|----------|-------------|
| `""`       | &lt;sha256&gt; | use the digest to fetch the image |
| `"latest"` | &lt;sha256&gt; | fetch the image tagged "latest", check that it has the digest given |

As it happens, the code here https://github.com/stefanprodan/timoni/pull/48/files#diff-54c697a1927e20930729129477a32e149e0285d47c7c05d2e60f72ad8395d595R144 resolves this by treating both `""` and `latest` as meaning " pull the latest, whatever the digest says". This is a bit confusing, but probably OK if it's explicitly called out in help docs.

Could it be better? If you didn't supply a default in the schema, you could distinguish between those cases, and distinguish between "pull the latest and check it has this digest", and "pull the latest, whatever it is". I can't see much use for that first one, but it is different to "just pull latest, whatever it is" -- perhaps you'd use it if you have automation to update the digest, and didn't want apply to proceed if that hasn't happened yet.

stefanprodan: > As it happens, the code here https://github.com/stefanprodan/timoni/pull/48/files#diff-54c697a1927e20930729129477a32e149e0285d47c7c05d2e60f72ad8395d595R144 resolves this by treating both "" (to which the default is applied) and latest (supplied explicitly) as meaning " pull the latest, whatever the digest says".

This is not the case, if the digest is specified and version is empty/latest, it will use the digest and not look at the latest digest. The `latest` value has a special meaning to Timoni as it does to Docker, and is ignored when coupled with digest. The digest check is only performed when semver version is set.

stefanprodan: I've added a test to prove this, "creates instance for digest ignoring latest".

squaremo: Owps, yes I got the explanation of that code the wrong way around, sorry! It's still a special case, but now I understand what it's about -- docker pull will treat e.g., `docker pull nginx:latest@sha256:<some other SHA> as though `latest` wasn't there.

cmd/timoni/bundle_apply.go
```diff
@@ -222,11 +169,16 @@ func applyBundleInstance(name, namespace, moduleURL, moduleVersion string, value
 		return err
 	}
 
+	if instance.Module.Digest != "" && mod.Digest != instance.Module.Digest {
+		return fmt.Errorf("the upstream digest of version %s doesn't match the specified digest %s",
```
hiddeco, line 173: Maybe mention the digest here of upstream which causes the mismatch?

docs/index.md
```diff
@@ -152,6 +152,17 @@ bundle: {
 }
 ```
 
+For deterministic operations, it is possible to pin a module version by its OCI digest:
```
hiddeco, line 155: Maybe include an example here of version and digest?

cmd/timoni/bundle_apply.go
stefanprodan: Done! Thanks

docs/index.md
stefanprodan: This level of detail would require a doc dedicated to bundles, where we explain the behaviour of combining version with digest. Going to make an issue for documenting bundles.

hiddeco (approved):
Looks like it does what it's documented to do, and happy with the feedback being taken into account 🎉

Issue #45, opened 2023-04-02
Support `stdin` for values files and bundles


stefanprodan:
For a better UX, Timoni should support reading from stdin using the `-` convention for:
- `timoni apply -f-`
- `timoni bundle apply -f-`

The implementation can be similar to https://github.com/fluxcd/flux2/blob/d4cc45f55e74b4517823b31836d3c5671c2be826/cmd/flux/push_artifact.go#L155-L162

trevex:
Just stumbled upon this project yesterday. As this is a good first issue and I was browsing the code, let me try to create a PR for this.

PR #49, opened 2023-04-03
Add stdin-support for values files and bundles

trevex:
Closes https://github.com/stefanprodan/timoni/issues/45


stefanprodan (commented):
@trevex this looks great thank you! Can you please add an example for stdin to both commands in the help section, something like `cat bundle.cue | timoni bundle apply -f-`.

trevex:
@stefanprodan, of course.  Help sections should now include an example as well.

stefanprodan (approved):
LGTM

Thanks @trevex 🥇

PR #51, opened 2023-04-03
Values from yaml and json files

### LINKED ISSUE #9: Allow for values to be supplied in other formats than cue
Currently Timoni expects that end-users who consume modules, to provider their custom values as `.cue` files. To easy adoption we should consider offering YAML and JSON as alternative formats. We could programmatically perform a cue import on the input YAML and JSON files in the CLI, so that Timoni's engine would not have to deal with different formats.

squaremo:
Aim: let people supply YAML and JSON files as values (i.e., fixes #9).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,6 @@ lint-samples: build
 	./bin/timoni mod lint ./examples/podinfo
 	cue fmt ./examples/podinfo-values/
 	./bin/timoni mod lint ./cmd/timoni/testdata/module
-	cue fmt ./cmd/timoni/testdata/module-values/
 	./bin/timoni mod lint ./internal/engine/testdata/module
 	cue fmt ./internal/engine/testdata/module-values
 
--- a/cmd/timoni/apply.go
+++ b/cmd/timoni/apply.go
@@ -19,7 +19,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -176,20 +175,11 @@ func runApplyCmd(cmd *cobra.Command, args []string) error {
 	logger.Printf("using module %s version %s", mod.Name, mod.Version)
 
 	if len(applyArgs.valuesFiles) > 0 {
-		for i, valuesFile := range applyArgs.valuesFiles {
-			if valuesFile == "-" {
-				path, err := saveReaderToFile(cmd.InOrStdin())
-				if err != nil {
-					return err
-				}
-
-				defer os.Remove(path)
-
-				applyArgs.valuesFiles[i] = path
-			}
+		valuesCue, err := convertToCue(cmd, applyArgs.valuesFiles)
+		if err != nil {
+			return err
 		}
-
-		err = builder.MergeValuesFile(applyArgs.valuesFiles)
+		err = builder.MergeValuesFile(valuesCue)
 		if err != nil {
 			return err
 		}
@@ -361,19 +351,4 @@ func runApplyCmd(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
-}
-
-func saveReaderToFile(reader io.Reader) (string, error) {
-	f, err := os.CreateTemp("", "*.cue")
-	if err != nil {
-		return "", fmt.Errorf("unable to create temp dir for stdin")
-	}
-
-	defer f.Close()
-
-	if _, err := io.Copy(f, reader); err != nil {
-		return "", fmt.Errorf("error writing stdin to file: %w", err)
-	}
-
-	return f.Name(), nil
 }
--- a/cmd/timoni/build.go
+++ b/cmd/timoni/build.go
@@ -24,7 +24,11 @@ import (
 	"path/filepath"
 	"strings"
 
+	"cuelang.org/go/cue/ast"
 	"cuelang.org/go/cue/cuecontext"
+	"cuelang.org/go/cue/format"
+	cuejson "cuelang.org/go/encoding/json"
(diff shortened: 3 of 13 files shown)


squaremo:
Notes for self: `cue` `eval`, `vet`, and `import` all accept YAML and JSON files. These all use some unexported helpers in https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/common.go. `cue import` has a lot of complication because it lets you import bits of filesystem, using various flags and command-line syntax, and similarly `cue vet`. `cue eval` (https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/eval.go) might have the most direct use of the unexported helpers, since it just produces the data value in the case of YAML and JSON.

squaremo:
`cue import` does this, essentially:

 - after a bit of flag massaging, it calls the unexported [`parseArgs`](https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/import.go#L269) to get a buildPlan
   - `parseArgs` calls [`loadFromArgs`](https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/common.go#L111) in the same file, which calls `load.Instances`, which does most of the work of looking for files to load, resolving dependencies, and so on, most of which we don't need, and ultimately,
     - calls [`filetypes.ParseArgs`](https://github.com/cue-lang/cue/blob/42580d5b27b9b66fe1d5cfed5e3ad4a3532c3c36/internal/filetypes/filetypes.go#L160) to interpret the command-line arguments as file names with intermingled "specs", and get a slice of `*build.File` for the next step. We don't need the specs as they are presented here, but it might be useful to allow e.g., `-f yaml:file.data`, so the nearby func `filetypes.ParseFile`might be useful for getting a `*build.File`
     - calls `cueFilesPackage`, which creates a `newFileProcessor` and [`fp.add`](https://github.com/cue-lang/cue/blob/42580d5b27b9b66fe1d5cfed5e3ad4a3532c3c36/cue/load/loader_common.go#L170)s each file -- that does a bit of preprocessing, which for non-CUE files, just means sticking it in the instance's `.OrphanedFiles`
   - [parseArgs calls `getDecoders`](https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/common.go#L571) on the instance, which goes through to orphaned files and tries to sort schemas from values. We are expecting only values, so a lot of this is irrelevant. For a value, it bundles the file up with a [decoder it constructs](https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/common.go#L481). The values end up being stuck in `buildPlan.orphans`, then processed by `placeOrphans`. This func is .. quite opaque to me .. but it calls decoder.File() to get an `ast.File` from the decoder, then it ends up doing a bunch more processing on each of those before sticking it in `buildPlan.imported`. It's not clear which bits of the processing are necessary for our purposes.
 - go through each b.imported and use format.Node() to serialise it to bytes, then write those to the exported file (we don't care about hoisting, because we don't expect embedded serialised values)

So, for an individual file, to get a data value (or convert it to a Cue file which will then be loaded), we'll need to do the following:
 - call [`filetypes.ParseFile`](https://github.com/cue-lang/cue/blob/42580d5b27b9b66fe1d5cfed5e3ad4a3532c3c36/internal/filetypes/filetypes.go#L226) with the filename argument, to get a `*build.File`;
 - construct a decoder, and use `decoder.File()` (and possibly some other bookkeeping) to get an `ast.File`;
 - use `format.Node` to serialise it out to a Cue file.

stefanprodan:
@squaremo maybe we could try using `cue/load` as suggested in https://github.com/stefanprodan/timoni/issues/9#issuecomment-1492987794

squaremo:
> maybe we could try using `cue/load`

Yeah, I was looking at that. It does a _lot_ of work, because it expects to interpret arguments given on the command line, and deal with things like schemas and packages; then you have to fish out the bits you wanted from the `Instance` at the end and work on them further. (e.g., the example given in that comment still has to go parse the data files). As an entry point it seemed a bit too all-encompassing for loading a single file with a known format.

I found a simple way, which is to use the packages under `cuelang.org/go/encoding` directly. There's both cuelang.org/go/internal/encoding and cuelang.org/go/encoding; judging by the comment made in [internal/internal.go](https://github.com/cue-lang/cue/blob/master/internal/internal.go#L17), it's a transitional thing.

stefanprodan:
Oh tests are failing due to running vet on testdata, please remove that, that was silly of me.

squaremo:
The writing out to files in $TMP is a bit um, temporary -- @stefanprodan what do you think of providing either `[]byte` or `func() ast.Value` to the engine rather than file paths?

stefanprodan (commented):
This looks ready good, thanks @squaremo. Could you please add the same function to the `timoni apply` command along with an example in the cmd help.

stefanprodan:
> what do you think of providing either []byte or func() ast.Value to the engine rather than file paths?

Yes I think `[]byte` would be great. Also maybe the yaml/json converter should move to the engine in a helper function that could be shared between the build and apply commands.

squaremo:
> Also maybe the yaml/json converter should move to the engine in a helper function that could be shared between the build and apply commands.

Oh yes, I forgot there would be other commands that use values files. OK, hold off while I tidy things up a bit ...

errordeveloper:
Just looking a this in an attempt to figure out how it all works, and what is a bit unclear to me is how validation would work. At a glance, it does seem a little bit complicated end.

I have previously implemented something similar with `FillPath`, which actually can accept `inteface{}`.

https://github.com/errordeveloper/kuegen/blob/a07f995a43d81cb98fe23685ebf3d499ffde2365/main.go#L139-L150

PR #72, opened 2023-04-11
Add a bundle example to quickstart

stefanprodan:
Recommend bundles for production use and add an example to the quickstart guide.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -150,13 +150,10 @@ To apply the above bundle on a cluster, save the file as `podinfo.bundle.cue` an
 timoni bundle apply -f podinfo.bundle.cue
 ```
 
-To list the instances created from the bundle:
+To delete all the instances defined in a bundle:
 
-```console
-$ timoni list -n podinfo
-NAME    MODULE                                          VERSION LAST APPLIED         
-podinfo oci://ghcr.io/stefanprodan/modules/podinfo      6.3.5   2023-04-10T16:20:07Z    
-redis   oci://ghcr.io/stefanprodan/modules/redis        7.0.10  2023-04-10T16:20:00Z
+```shell
+timoni bundle delete -f podinfo.bundle.cue
 ```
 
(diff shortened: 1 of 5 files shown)


docs/quickstart.md
```diff
@@ -107,3 +107,56 @@ To uninstall an instance and delete all the managed Kubernetes resources:
...
+}
+```
+
+To apply the above bundle on a cluster, save the file as `podinfo.bundle.cue` and run:
+
+```shell
+timoni bundle apply -f podinfo.bundle.cue
+```
+
+To list the instances created from the bundle:
```
squaremo, line 153: This is a bit of a misdirection, because the text says "list things created from the bundle" but the command does that only by coincidence -- both instances happen to be in the namespace mentioned, and evidently nothing else is). This example command doesn't generalise. What if there were instances from some other bundle in that namespace? or, the bundle deployed to more than one namespace?

I don't think this is a serious problem but it might give people the wrong idea, or confuse them. Perhaps the text could be "To list the instances in the namespace used by the bundle:" or similar.

squaremo (approved):
Lucid and reasonable :+1:

docs/quickstart.md
hiddeco: Made a likewise comment to Stefan in private, and we both had the same ideas around options to improve this beyond phrasing:

1. Make defining a name in a bundle required, which can then be used as a label on the backing Secret. This could then be used to filter the output of the command based on this label.
2. Allow supplying a bundle file, which is then used to filter the output. But this seems to be the lesser option, given not every user may have access to the source (or is a UI).

stefanprodan: A bundle can contain instances in different namespaces, if we go with labeling, to find them all would be something like:

``
timoni ls --bundle=my-bundle --all-namespaces
``

stefanprodan: I have replaced listing with deletion in the quickstart. I'll add back listing once we have the label in place. Thanks @squaremo and @hiddeco for the feedback!

PR #87, opened 2023-04-19
Improve how `GetResource` works

### LINKED ISSUE #84: yaml: unsupported node string (*ast.Ident)
I got this error while porting and existing CUE project from [`kuegen`](https://github.com/errordeveloper/kuegen) to Timoni:

```
✗ failed to extract objects, error: encoding objects for all failed, error: yaml: unsupported node string (*ast.Ident)
```

When I ran `kuegen`, I got a different error, which was a bit more meaningful:

```
main.go:248: cue: marshal error: template.items.4.subjects.0.namespace: cannot convert incomplete value "string" to JSON
```

It looks like marshaling logic might be implemented differently in Timoni, I can take a look into that.

### LINKED ISSUE #86: open list yields `error: objects in foo are not of type cue.ListKind, got struct`
Seems like the type check is wrong. 

(Hunch: possibly related to #84, i.e. something is missing on export).

errordeveloper:
- (fix #86) drop unnecessary type check, as for some reason it doesn't work for open list; instead `List()` call that is already used will report an error when the underlying value is not a list
- (fix #84) ensure all values are final and handle error that was previously ignored, which should report incomplete types

### DIFF
--- a/internal/engine/resourceset.go
+++ b/internal/engine/resourceset.go
@@ -18,9 +18,10 @@ package engine
 
 import (
 	"bytes"
+	"fmt"
+
 	"cuelang.org/go/cue"
 	"cuelang.org/go/encoding/yaml"
-	"fmt"
 	"github.com/fluxcd/pkg/ssa"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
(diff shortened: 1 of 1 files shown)


errordeveloper:
cc @squaremo 

This probably needs some unit tests.

stefanprodan (approved):
LGTM

Thanks @errordeveloper 🏅

internal/engine/resourceset.go
```diff
@@ -39,34 +40,36 @@ type ResourceSet struct {
...
 		name := iter.Selector().String()
 		expr := iter.Value()
-		switch expr.Kind() {
-		case cue.ListKind:
-			items, err := expr.List()
-			if err != nil {
-				return nil, fmt.Errorf("listing objects for %s failed, error: %w", name, err)
-			}
+		if expr.Err() != nil {
+			return nil, fmt.Errorf("getting value of resource list %q failed, error: %w", name, expr.Err())
```
hiddeco, line 51: Not related to this PR, but something that I noticed while going over this. What's the benefit of suffixing with `, error:` when doing error wrapping? Given it does not make much sense for the individual error without the context from the stack of wrapped errors.

cc: @stefanprodan

stefanprodan: It's just my bad habit..  should be `failed: %w`?

hiddeco: Yes, because then if you "detach" the error from the stack it looks like a thing on it's own without giving you the idea something is missing :smile: 

> getting value of resource list %q failed

Issue #83, opened 2023-04-18
submodules


errordeveloper:
When working with an existing CUE project, Timoni doesn't recognise the top-most `cue.mod` dir as it just copies the module directory to a `/tmp`, and thereby isolates it from the parent project.

errordeveloper:
To work around this, one can place Timoni configs in the top-level dir and import what needs importing from the existing subpackage(s).

stefanprodan:
Fixed in https://github.com/stefanprodan/timoni/pull/602 modules are no longer copied in tmp

PR #90, opened 2023-04-25
Optimise copying of modules

### LINKED ISSUE #80: improve handling of existing projects
One challenge with using Timoni in an existing project right now is that if top-level dir is used, it will end up copying various dirs like `vendor` and `.git`, which can be _very slow_ and the recursive copier method is currently serial.
It probably makes sense to speed up the process, but reading `.gitignore` and/or having `.timoniignore` would also help. Aside from that, due to #55 there is no way for a new user to figure out what's going on without digging into the code.

errordeveloper:
- only copy `*.cue` files, as CUE cannot evaluate or read any other files
- perform I/O operations in parallel
- skip commonly used dirs that are dimmed irrelevant (`.git`, `vendor` and `$GOPATH`)

towards #80


internal/engine/fetcher.go
```diff
@@ -80,7 +80,24 @@ func (f *Fetcher) Fetch() (*apiv1.ModuleReference, error) {
 		Digest:     "unknown",
 	}
 
-	return &mr, CopyModule(f.src, modulePath)
+	copier, err := NewCopier(f.src, modulePath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create copier, error: %w", err)
+	}
+
+	copier.AddSkipDirs(".git", "vendor")
```
errordeveloper, line 88: I suppose this is not very controversial.

internal/engine/fetcher.go
```diff
@@ -80,7 +80,24 @@ func (f *Fetcher) Fetch() (*apiv1.ModuleReference, error) {
...
+	if gopath, haveGoapth := os.LookupEnv("GOPATH"); haveGoapth {
+		copier.AddSkipDirs(gopath)
```
errordeveloper, line 90: However, this one might be a bit controversial, as env lookup kind of changes things. Making explicit config with `.timoniignore` or `.gitignore` might be more sensible, but on the other hand, I'm still not sure why files aren't read in the original location.

errordeveloper:
@stefanprodan @squaremo I've marked this as a draft because I'm not entirely convinced this is the best route (as opposed to just not copying any files). Let's discuss that first. If we must copy, then something like this is gonna be needed.

internal/engine/fetcher.go
stefanprodan: Modules come from remote registries, I see no way around using some tmp dir. The situation when then module is at a local path is only during development.

errordeveloper: Perhaps this fetcher piece should be an interface with two implementations - local dir and OCI?
To me it looks like the use case for local dir is that it is used in development and wherever during build in CI.

errordeveloper:
If we split the fetcher piece into two implementations - local dir and OCI, we will avoid #83 and the slowness that `.git`, `vendor`, and `$GOPATH` in sub dir can cause.

errordeveloper:
> We'll use Flux's `.sourceignore` and it's ignore package that does the right thing. I'll create an issue for this.

I'm not fully convinced Timoni has to require the user to add another file. I think it should be possible to just collect all visible CUE sources, but exclude ones that use `if(@<tag>)`. It's possible a file may be needed in some cases, but perhaps it could be entirely optional.

I'm still mostly in favour of the option where local dir vs published OCI images are treated differently when it comes to loading directly vs loading from a copy in temp dir.

stefanprodan:
> I'm not fully convinced Timoni has to require the user to add another file. I think it should be possible to just collect all visible CUE sources

I don't think this would be Ok for `timoni mod push`, we need to let people decide what gets published, for example, Timoni expect a README.md to be present in the module root, if we include only .cue files than this would break the assumption that a published module contains the values docs.

errordeveloper:
> I don't think this would be Ok for `timoni mod push`, we need to let people decide what gets published, for example, Timoni expect a README.md to be present in the module root, if we include only .cue files than this would break the assumption that a published module contains the values docs.

Sure, but you can do that by including all `.cue` files plus `README.md`! Just saying ;)

stefanprodan:
A `README.md` could refer to other assets like a png with a logo or diagrams, if we exclude pngs we'll break the readme.

errordeveloper:
> A `README.md` could refer to other assets like a png with a logo or diagrams, if we exclude pngs we'll break the readme.

A markdown parser could tell you what those are, but that might be an overkill 😸 

One alternative to using a "traditional" ignore file, would be to have something like:
```CUE
timoni: {
  docs: {
    readme: *"README.md" | string | bytes
    logo?: string | bytes
  }
}
```
This would allow for more flexibility, and make the API fully define in CUE. We can debate if this lives in `main` package or a separate thing (like under `cue.mod` or something), but overall this does seem rather neat to me.

Issue #76, opened 2023-04-16
Track instances installed from a bundle


stefanprodan:
We need a way to track which instances installed on a cluster come from which bundle. The proposed solution is to label the instances with the bundle name, then allow users to query the cluster API based on that name.

Tasks:
- [x] Add a required filed called `name` to the Bundle API 
- [x] At apply time, label the instances storage (Kubernetes Secret) with `bundle.timoni.sh/name: <name>`
- [x] Add `--bundle=<name>` to the `timoni ls -A` to list all instances on a cluster filtered by bundle name

Should we?
- Issue a warning if an instance is managed by a different bundle before overriding the label value
- Add the bundle name as a column to `timoni ls`
- Add `--name` to `timoni delete bundle` as an alternative to `-f`

stefanprodan:
Implemented in https://github.com/stefanprodan/timoni/pull/91

PR #91, opened 2023-04-29
feat: enable name-oriented bundle management

siiimooon:
The PR contains a take on https://github.com/stefanprodan/timoni/issues/76. Implementing both name-oriented Bundle deletion and Bundle-aware Instance listing, by extending the Bundle API with `.metadata.name` and tracking Instances by label `bundle.timoni.sh/name`.

Test coverage, documentation, CLI adjustments, etc, remains. Can fix this if the overall approach is okay.

### DIFF
--- a/api/v1alpha1/bundle.go
+++ b/api/v1alpha1/bundle.go
@@ -20,6 +20,9 @@ const (
 	// BundleAPIVersionSelector is the CUE path for the Timoni's bundle API version.
 	BundleAPIVersionSelector Selector = "bundle.apiVersion"
 
+	// BundleName is the CUE path for the Timoni's bundle name.
+	BundleName Selector = "bundle.name"
+
 	// BundleInstancesSelector is the CUE path for the Timoni's bundle instances.
 	BundleInstancesSelector Selector = "bundle.instances"
 
@@ -37,6 +40,9 @@ const (
 
 	// BundleValuesSelector is the CUE path for the Timoni's bundle instance values.
 	BundleValuesSelector Selector = "values"
+
+	// BundleNameLabelKey is the Kubernetes label key for tracking Timoni's bundle by name.
+	BundleNameLabelKey = "bundle.timoni.sh/name"
 )
 
 // BundleSchema defines the v1alpha1 CUE schema for Timoni's bundle API.
@@ -46,6 +52,7 @@ import "strings"
 
 #Bundle: {
 	apiVersion: string & =~"^v1alpha1$"
+	name: string & =~"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$" & strings.MaxRunes(63) & strings.MinRunes(1)
 	instances: [string & =~"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$" & strings.MaxRunes(63) & strings.MinRunes(1)]: {
 		module: close({
 			url:     string & =~"^oci://.*$"
--- a/cmd/timoni/apply.go
+++ b/cmd/timoni/apply.go
@@ -92,16 +92,17 @@ The apply command performs the following steps:
 }
 
 type applyFlags struct {
-	name        string
-	module      string
-	version     flags.Version
-	pkg         flags.Package
-	valuesFiles []string
-	dryrun      bool
-	diff        bool
-	wait        bool
-	force       bool
-	creds       flags.Credentials
+	name               string
+	module             string
+	version            flags.Version
+	pkg                flags.Package
+	valuesFiles        []string
+	dryrun             bool
+	diff               bool
+	wait               bool
+	force              bool
+	overwriteOwnership bool
+	creds              flags.Credentials
 }
 
 var applyArgs applyFlags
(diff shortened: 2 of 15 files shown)


siiimooon:
Made some minor improvements - including conflict-aware apply.

From the issue:
`Issue a warning if an instance is managed by a different bundle before overriding the label value`

As of the PR, one would have to apply with force in order to gain ownership of an existing instance.

cmd/timoni/apply.go
```diff
@@ -231,10 +231,18 @@ func runApplyCmd(cmd *cobra.Command, args []string) error {
 
 	exists := false
 	sm := runtime.NewStorageManager(rm)
-	if _, err := sm.Get(ctx, applyArgs.name, *kubeconfigArgs.Namespace); err == nil {
+	instance, err := sm.Get(ctx, applyArgs.name, *kubeconfigArgs.Namespace)
+	if err == nil {
 		exists = true
 	}
```
stefanprodan, line 237: Wouldn't this break upgrades as `instancesOwnershipConflicts` doesn't check the bundle name against the instance label?

siiimooon: Not sure if I understand the question.

By _instance label_ I assume you refer to the Bundle tracking label of the Instance(?). Instances deployed outside Bundles are, as of now, expected to have no Bundle tracking label. `instancesOwnershipConflicts` only checks for the presence of any Bundle tracking label on the Instance.

stefanprodan: Sorry I got confused, you're right we should warn people they are are overriding an instance managed by a bundle. Can we tell them to use `--force` in the error message, if they really want to override?

siiimooon: Added a new commit, where ownership transfer has been moved to a separate flag. I guess it was a bit unfortunate to combine this with `--force`, as force apply has other significant consequences.

docs/bundles.md
```diff
@@ -104,9 +105,9 @@ List the instances in the `podinfo` namespace:
 === "output"
 
      ```text
-     NAME    MODULE                                          VERSION LAST APPLIED         
-     podinfo oci://ghcr.io/stefanprodan/modules/podinfo      6.3.5   2023-04-10T16:20:07Z    
-     redis   oci://ghcr.io/stefanprodan/modules/redis        7.0.10  2023-04-10T16:20:00Z
+     NAME    NAMESPACE         MODULE                                          VERSION LAST APPLIED          BUNDLE
```
stefanprodan, line 108: Can you please add above `timoni list --bundle podinfo -A`

docs/bundles.md
```diff
@@ -311,6 +314,20 @@ timoni bundle apply --force -f bundle.cue
...
+will fail. To transfer ownership to the current Bundle, you need to set the `--overwrite-ownership` flag.
+
+Example:
+
+```shell
+timoni bundle apply --overwrite-ownership -f bundle.cue
+```
+
+With `--force`, Timoni will recreate only the resources that contain changes
+to immutable fields.
```
stefanprodan, line 329:
```suggestion

```

docs/bundles.md
```diff
@@ -311,6 +314,20 @@ timoni bundle apply --force -f bundle.cue
 With `--force`, Timoni will recreate only the resources that contain changes
 to immutable fields.
 
+### Overwrite ownership
```
stefanprodan, line 317:
```suggestion
### Transfer ownership
```

internal/engine/bundle_builder.go
```diff
@@ -33,7 +33,13 @@ type BundleBuilder struct {
 	files   []string
 }
 
+type Bundle struct {
+	Name      string
+	Instances []BundleInstance
+}
+
 type BundleInstance struct {
+	Bundle    *Bundle
```
stefanprodan, line 42: This creates a cyclic dependency between the two type, could we have `Bundle` in  `BundleInstance` as type string?

docs/bundles.md
siiimooon: Added a dedicated example for name-oriented listing, rather than replacing the namespace-oriented one. Can replace the latter if you'd like.

api/v1alpha1/bundle.go
```diff
@@ -20,6 +20,9 @@ const (
 	// BundleAPIVersionSelector is the CUE path for the Timoni's bundle API version.
 	BundleAPIVersionSelector Selector = "bundle.apiVersion"
 
+	// BundleMetadataName is the CUE path for the Timoni's bundle name.
+	BundleMetadataName Selector = "bundle.metadata.name"
```
stefanprodan, line 24: Given that Bundles are not Kubernetes CRDs and they don't have `.spec`, I think it's best to remove `.metadata` and have `name` as top level prop.

hiddeco: Vouch for this as well, to prevent a Kustomize-like (and eksctl, etc.) scenario where people get confused about things looking like a Kubernetes API object, while they actually aren't.

siiimooon:
> LGTM
> 
> Thanks @siiimooon awesome contribution 🏅 🎖️ 🥇

Thanks to you guys too, for creating the next-gen of Kubernetes manifest management. Would you like me to squash the commits now, or will you do this at the time of merge?

stefanprodan:
@siiimooon if you could redo the commit message to fit the 72 characters limit would be great

hiddeco:
> if you could redo the commit message to fit the 72 characters limit would be great

50 characters for the first sentence, 72 line limit for the description :-)

Issue #57, opened 2023-04-05
Programmatic vetting of modules


stefanprodan:
Currently the `timoni mod lint` command shell-out to the `cue` CLI to perform `cue vet -c`. Instead of depending on the cue CLI, the lint command should programmatically vet the module, this is already done in `timoni build`, so we could extract the build logic and expand the validation errors in the lint output.

PR #93, opened 2023-05-03
Introduce instance schema and module linting

stefanprodan:
Changes:
- Add instance schema to API
- Enforce instance schema validation for modules
- Implement programmatic vetting of modules (fix: #57)


hiddeco (approved):
Yay for getting rid of `, error:` :tada:

Issue #97, opened 2023-05-07
Support for building instances of a bundle


gustavclausen:
I'm looking to build all instances of a bundle and print the resulting Kubernetes resources – similar to how a single instance can be built using the [`timoni build`](https://timoni.sh/cmd/timoni_build/) command.

My current solution is to run an external script that parses the bundle file and execute `timoni build` for each item in the `bundle.instances` property. But I'm thinking that a `timoni bundle build` command could be added to perform this.

I'm open to implement it if you're giving the green light to this.

stefanprodan:
@gustavclausen I was planing to add an issue for this 😄 Looking forward to your PR.

PS. The build command would have same code as apply dry-run, but instead of applying the manifests, it would write them to stdout. 

Thanks!

PR #103, opened 2023-05-13
Add support for building instances of a bundle

gustavclausen:
This adds support for building all instances of a bundle and print the resulting Kubernetes resources to stdout using the following command: `timoni bundle build`. This is intended to be work similar to how a single instance can be built using the [timoni build](https://timoni.sh/cmd/timoni_build/) command.

Fix: #97.


stefanprodan (approved):
LGTM

Thanks @gustavclausen awesome contribution 🥇

Issue #55, opened 2023-04-03
Arrange proper logging mechanisms


hiddeco:
With features like #25, there is a growing need for a logger that can be used not just in `cmd/timoni`, but also in underlying libraries.

We have had positive experiences with [`github.com/go-logr/logr`](https://github.com/go-logr/logr) as a logging API in controller-runtime based projects. Specifically because of its [context logger](https://github.com/go-logr/logr/blob/82f95559d383fe44d114f0a65582ecc3df8a0f66/logr.go#L403-L437) capabilities. Which saves you from passing around an additional logger value throughout the code-base, while stimulating the usage of contexts (for longer running actions).

As `logr` by itself does not offer any logging capabilities, we need to decide on an [implementation to use](https://github.com/go-logr/logr#implementations-non-exhaustive). Preferably one which has pretty, console friendly, human readable output (as our main goal is to be a CLI), while also allowing it to be e.g. JSON.

 My personal preference would be something lighter than `go.uber.org/zap` ([`zapr`](https://github.com/go-logr/zapr)). For example `github.com/rs/zerolog`, which also has a [`ConsoleWriter`](https://github.com/rs/zerolog#pretty-logging).

Stefan and I briefly evaluated [`github.com/pterm/pterm`](https://github.com/pterm/pterm) as an option (which happens to have a [`logr` implementation](https://github.com/ccremer/plogr)), but had concerns about this being console only, and less friendly to machines (or potential log capturing CI runners). In addition, using the implementation would not provide access to many of the goodies the library itself has, making it much less powerful and therefore a less ideal option. For the spinner capabilities in the CLI, we could instead use [`github.com/briandowns/spinner`](https://github.com/briandowns/spinner) in `cmd/timoni`.

As part of the arrangement, `timoni` should get flags to control the log level and output type.

PR #112, opened 2023-05-26
Add pretty print for logs and diff output

stefanprodan:
Changes:
- Implement contextual logging with logr and zerolog (fix: #55)
- Implement colorised Kubernetes objects diffing with dyff
- Add a global flag for disabling the colorised output and log timestamps `--log-pretty=false`
- Add spinner before starting long-running operations


stefanprodan:
With logr you need to format  the message with `Sprintf` while zerolog it’s just `Msgf`.

hiddeco:
Yes, [with reason](https://github.com/go-logr/logr#why-not-allow-format-strings-too).

The downside of the current approach is that you have no contextual logging, nor are proving (potential) SDK users with the promise of `logr` in which they can inject their own implementation without worrying.

PR #120, opened 2023-05-31
Add support for injecting values from environment variables

stefanprodan:
This PR introduces a CUE attribute in the format `@timoni(env:[string|number|bool]:[ENV_VAR_NAME])`. The attribute can be placed in bundles next to any field to set its value from the environment at runtime.

### Motivation

As shown [here](https://github.com/stefanprodan/timoni/tree/main/examples/bundles), a Timoni bundle can be made out of multiple files, so we can have a CUE file that contains some values generated at runtime from various sources e.g. an encrypted store, CI vars, etc. While this approach works, it forces uses into storing or generating CUE files in CI, which can be cumbersome. By introducing support for environment variables, users can easily inject values without having to modify their pipelines, as most workflow engines can set env vars before running a command. 

Another reason for implementing environment variables is offering a similar feature to Helm's `--set-file` and `--set-json`. Some modules may require users to provide TLS certs, or app config files (JSON, INI, XML). Instead of forcing users into embedding the file text content in a bundle, the content can be stored in an env var and Timoni will inject it at runtime in the bundle as  multi-line string, preserving the original indentation.

An alternative to using the environment to inject file contents, would be adding an attribute for reading files from disk e.g. `@timoni(import:string:ssh/id_ecdsa.pub)`. But then we'll have to deal with relative paths (to the bundle or to the execution root?), symlinks, path traversal protection, etc. I think for now, we can ship the environment solution and see what issues users will have with it, then decide if we really need a file reader.

### Example

```cue
appName: string @timoni(env:string:APP_NAME)

bundle: {
	apiVersion: "v1alpha1"
	name:       "\(appName)"
	instances: {
		"\(appName)": {
			module: {
				url:     "oci://my-registry/modules/app"
				version: "1.0.0" @timoni(env:string:APP_VER)
			}
			values: {
				// required string (can be multi-line)
				sshKey: string @timoni(env:string:SSH_KEY)
				// optional boolean (defaults to false)
				isAdmin: false @timoni(env:bool:IS_ADMIN)
				// required number
				age: int @timoni(env:number:AGE)
			}
		}
	}
}
```

Export the env vars and run the `timoni bundle apply` command.

```shell
EXPORT APP_NAME="my-app"
EXPORT SSH_KEY=$(cat .ssh/id_ecdsa.pub)
EXPORT IS_ADMIN="true"
EXPORT AGE="41"

timoni bundle apply -f bundle.cue
```

At apply time, Timoni injects the fields values from the environment, if a specified env var is not found and if a default is not provided, the build operation with fail with an `incomplete value` error.

### DIFF
--- a/cmd/timoni/bundle_apply.go
+++ b/cmd/timoni/bundle_apply.go
@@ -92,16 +92,7 @@ func init() {
 }
 
 func runBundleApplyCmd(cmd *cobra.Command, _ []string) error {
-	bundleSchema, err := os.CreateTemp("", "schema.*.cue")
-	if err != nil {
-		return err
-	}
-	defer os.Remove(bundleSchema.Name())
-	if _, err := bundleSchema.WriteString(apiv1.BundleSchema); err != nil {
-		return err
-	}
-
-	files := append(bundleApplyArgs.files, bundleSchema.Name())
+	files := bundleApplyArgs.files
 	for i, file := range files {
 		if file == "-" {
 			path, err := saveReaderToFile(cmd.InOrStdin())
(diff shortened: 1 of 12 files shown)


squaremo:
As a meta-comment: I really like the style of conversational, but carefully considered, rationale given here.

Issue #74, opened 2023-04-12
[proposal] Encryption of Timoni values


phoban01:
# Context

The following is a proposal for supporting the encryption of Timoni values based on prior art (see: https://github.com/phoban01/cue-sops). 

The proposal was first mooted in #71.

# Proposal

Timoni enables the encryption of sensitive values via built-in SOPS support.

Given the following config definition:

```cue
#Config: {
  api: {
    url:  "https//api.github.com/user"
    token: string
  }
}

```

A values file providing an API key is marked as sensitive using the `@secret()` annotation:

```cue
# api-values.cue
values: {
    api: token: "gh_personalaccesstoken" @secret()
}
```

The plaintext value can then be encrypted in-place:

`timoni encrypt -f api-values.cue`

This will produce the following result:

```cue
# api-values.cue
values: {
	api: token: "ENC[AES256_GCM,data:0SeH+BIX6SwJBsgwLmDOJHU7,iv:Fx1bpRKrz4wKztuEXMfa0KuRqLcOu9ZLT8OYdH+i58c=,tag:IoDhNZpGnGhqmDllgUVdUg==,type:str]" @secret()
}

// DO NOT EDIT: auto-generated by timoni
sops: {
	kms:      null
	gcp_kms:  null
	azure_kv: null
	hc_vault: null
	age: [{
		recipient: "age1ethasxep4zkax64yfx35rn2t4yeul4254w764l9gtasvn2rwpv7s733dq7"
		enc: """
			-----BEGIN AGE ENCRYPTED FILE-----
			YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IFgyNTUxOSBQMk43bjFuUytFNTlIclNW
			Y1RnNWEwc2FGOUd4VW5NODNwdEVKOXlJd2k4ClNubDN0Qktuck5IVnN6ZjZBOTEz
			OFhNRUc3aUs1Y09DQTF6OTlWRU9ZQ00KLS0tIGdGTUlZWUJyVkZKZXdvMzZhV294
			c0E5bHVkSHc0MkhFUnhiODFlbzV5SE0KlNEhfwHl/VDZzfkpGb2/s7KbTFRA4U/K
			u5OM5P2YTvpSkmVbdVLLcX7eFHVyLZOukarFXEZ65rq9baMO0lJ3Vg==
			-----END AGE ENCRYPTED FILE-----

			"""
	}]
	lastmodified:    "2023-04-01T12:00:00Z"
[...]

PR #128, opened 2023-06-21
feat: add YAML and JSON support for bundles

b4nst:
This little addition add YAML support for bundle file. It can be useful for multiple purposes, like using `sops` encrypted files until we address #74.

I'm also referencing cue-lang/cue/discussions/2452 in case a better way of including Orphan files exists.

### DIFF
--- a/cmd/timoni/bundle_build_test.go
+++ b/cmd/timoni/bundle_build_test.go
@@ -31,36 +31,62 @@ func Test_BundleBuild(t *testing.T) {
 	))
 	g.Expect(err).ToNot(HaveOccurred())
 
-	bundleData := fmt.Sprintf(`
+	bundleCue := fmt.Sprintf(`
 appName: string @timoni(env:string:TEST_BBUILD_NAME)
 bundle: {
 	apiVersion: "v1alpha1"
-	name: "%[1]s"
+	name: string
 	instances: {
 		"\(appName)": {
 			module: {
-				url:     "oci://%[2]s"
-				version: "%[3]s"
+				url:     "oci://%[1]s"
+				version: "%[2]s"
 			}
-			namespace: "%[4]s"
+			namespace: "%[3]s"
 			values: server: enabled: false
 			values: domain: string @timoni(env:string:TEST_BBUILD_HOST)
 		}
 		backend: {
 			module: {
-				url:     "oci://%[2]s"
-				version: "%[3]s"
+				url:     "oci://%[1]s"
+				version: "%[2]s"
 			}
-			namespace: "%[4]s"
+			namespace: string
 			values: client: enabled: bool @timoni(env:bool:TEST_BBUILD_ENABLED)
 		}
 	}
 }
-`, bundleName, modURL, modVer, namespace)
+`, modURL, modVer, namespace)
 
-	bundlePath := filepath.Join(t.TempDir(), "bundle.cue")
-	err = os.WriteFile(bundlePath, []byte(bundleData), 0644)
-	g.Expect(err).ToNot(HaveOccurred())
+	bundleData := bundleCue + fmt.Sprintf(`
+bundle: name: "%[1]s"
+bundle: instances: backend: namespace: "%[2]s"
+`, bundleName, namespace)
+
+	bundleJson := fmt.Sprintf(`
+{
+	"bundle": {
+		"name": "%[1]s"
+	}
+}
+`, bundleName)
+
+	bundleYaml := fmt.Sprintf(`
+bundle:
+  instances:
+    backend:
+      namespace: %[1]s
+`, namespace)
+
+	wd := t.TempDir()
+	cuePath := filepath.Join(wd, "bundle.cue")
+	g.Expect(os.WriteFile(cuePath, []byte(bundleCue), 0644)).ToNot(HaveOccurred())
+
+	yamlPath := filepath.Join(wd, "bundle.yaml")
+	g.Expect(os.WriteFile(yamlPath, []byte(bundleYaml), 0644)).ToNot(HaveOccurred())
+
+	jsonPath := filepath.Join(wd, "bundle.json")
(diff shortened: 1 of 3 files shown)


stefanprodan (commented):
Hey @b4nst we have similar feature for [values](https://github.com/stefanprodan/timoni/blob/main/cmd/timoni/build.go#L205), where we allow both YAML and JSON. I think we need to support JSON as well for consistency.

Also a test with a bundle made out of 3 files CUE, YAML and JSON would be good.

b4nst:
Hey @stefanprodan, just added JSON support and a test case with the combo of all file types.

stefanprodan:
Thanks @b4nst can you please rebase with main, solve conflicts and force push to unblock CI.

b4nst:
Of course! Sorry I went a bit fast on the push 😅

b4nst:
Ok rebased, seems there's an issue with the Copy and YAML files during workspace creation introduced with the injection system (break the content of the file). I'll give it a look

stefanprodan:
@b4nst I suggest we put in the workspace not the jsons and yamls but their cue value.

b4nst:
Yup, that's what I started (moving the extract part into the workspace function). Incoming stoday

internal/engine/bundle_builder.go
```diff
@@ -69,23 +72,44 @@ func (b *BundleBuilder) InitWorkspace(workspace string) error {
...
 		if err != nil {
-			return fmt.Errorf("failed to inject %s: %w", fn, err)
+			return err
 		}
-
-		if err := os.WriteFile(f, data, os.ModePerm); err != nil {
-			return fmt.Errorf("failed to inject %s: %w", fn, err)
+		defer f.Close()
+
+		if a, err := yaml.Extract(file, f); err == nil {
```
stefanprodan, line 82: This approach has many issues, please let's look at the file extension in a switch statement instead of hiding errors. If a cue, json or yaml is invalid, we should tell users and exit, instead of silently ignoring errors which would result in applying something on a cluster that was never intended.

b4nst: Okay with me, but it means we don't support file with incorrect extension (which is probably a fair assumption).

stefanprodan: Yes we don't support incorrect extension, the switch should match:  https://github.com/stefanprodan/timoni/blob/main/cmd/timoni/build.go#L205

b4nst: Perfect, going to grab some food and pushing you a fix after

stefanprodan: Thanks

b4nst: Okay did a switch around extensions. Also refactored the injector so it harmonizes the pattern and avoid double parsing / writing of files. I also dropped the copy dependency in the meantime.

stefanprodan: Looks great, thanks a lot @b4nst.

I was wandering if we can create the workspace in-memory instead of on disk in tmp. I haven't figure out how to pass bytes to cue build, but it must be a way.

cmd/timoni/bundle_build_test.go
```diff
@@ -31,35 +30,66 @@ func Test_BundleBuild(t *testing.T) {
...
+`, bundleName)
+
+	bundleYaml := fmt.Sprintf(`
+bundle:
+  instances:
+    backend:
+      namespace: %[1]s
+`, namespace)
+
+	cuef, err := ioutil.TempFile(t.TempDir(), "*.cue")
```
hiddeco, line 80: `ioutil` is deprecated since Go 1.17, this should be `os.CreateTemp` from top of head.

stefanprodan: Good catch, should be like we do it elsewhere:

```
	bundlePath := filepath.Join(t.TempDir(), "bundle.cue")
	err = os.WriteFile(bundlePath, []byte(bundleData), 0644)
	g.Expect(err).ToNot(HaveOccurred())
```

`t.TempDir()` does a cleanup so we don't leave stuff in tmp.

b4nst: Oh nice catch! Will address it, thanks

internal/engine/bundle_builder.go
stefanprodan: New PR for sure, after we get this in shape and merged. Thanks for looking into it.

cmd/timoni/bundle_build_test.go
b4nst: Done

internal/engine/bundle_builder.go
b4nst: We might want to revive https://github.com/cue-lang/cue/issues/607 for a proper in-memory support

cmd/timoni/bundle_build_test.go
```diff
@@ -31,47 +31,72 @@ func Test_BundleBuild(t *testing.T) {
...
+`, bundleName)
+
+	bundleYaml := fmt.Sprintf(`
+bundle:
+  instances:
+    backend:
+      namespace: %[1]s
+`, namespace)
+
+	cuePath := filepath.Join(t.TempDir(), "bundle.cue")
```
hiddeco, line 81: Every time `t.TempDir()` is called, a new temporary directory is created from what I recall. I would therefore call it once and then re-use it.

hiddeco: > Each subsequent call to t.TempDir returns a unique director
> &mdash; https://pkg.go.dev/testing#T.TempDir

b4nst: Hum indeed. I don't think it will drastically improve perf but that would be cleaner indeed. Wilco

hiddeco (approved):
This looks good to me :100:

stefanprodan (approved):
LGTM

Thanks @b4nst 🥇

Issue #127, opened 2023-06-21
define schema for image values


errordeveloper:
It would be very handy if Timoni has defined a standard schema for images.

errordeveloper:
What is used in the podinfo example if already a good starting point:

```CUE
#Image: {
	repository: string
	tag:        string
	digest?:    string
	pullPolicy: *"IfNotPresent" | string
}
```

An iteration on this yields:

```CUE
#Image: {
	repository: string
	tag:        string
	digest?:    string
}

#ContainerImage: {
	#Image
	pullPolicy: *"IfNotPresent" | string
	pullSecrets?: [...string]
}
```

stefanprodan:
Could we make this easier, so people don't have to write 2 ifs for every container? 

```
if _config.image.digest == _|_ {
  image: "\(_config.image.repository):\(_config.image.tag)"
}
if _config.image.digest != _|_ {
  image: "\(_config.image.repository)@\(_config.image.digest)"
}
imagePullPolicy: _config.image.pullPolicy
```

errordeveloper:
I think it would be good to try tackling conditional digest setting logic. 

I something like this could work:
```CUE
#Image: {
	#repository: string
	#tag:        string
	#digest?:    string

	reference: string


	if #digest == _|_ {
		reference: "\(#repository):\(#tag)"
	}
       	if #digest != _|_ {
		reference: "\(#repository)@\(#digest)"
	}
}

exampleImage: #Image & {
	#repository: "example.com/image"
	#tag: "latest"
}

exampleCotainer: image: exampleImage.reference
```

https://cuelang.org/play/?id=4j7TI-NJrva#cue@export@cue

stefanprodan:
How would this look in #Config and in values.cue? Do users need to set it as `#tag` and `#digest`?

jilleJr:
> Could we make this easier, so people don't have to write 2 ifs for every container?
> 
> ```cue
> if _config.image.digest == _|_ {
>   image: "\(_config.image.repository):\(_config.image.tag)"
> }
> if _config.image.digest != _|_ {
>   image: "\(_config.image.repository)@\(_config.image.digest)"
> }
> imagePullPolicy: _config.image.pullPolicy
> ```

Would prefer if the `timoni mod init` bootstrap suggested something like this:

```cue
#ImageRef: {
	reference: string
	...
}

#ImageTag: {
	repository: string
	tag:        string
	reference:  *"\(repository):\(tag)" | string
	...
}

[...]

PR #132, opened 2023-07-07
Introduce experimental cue schema for OCI images

stefanprodan:
This is an experiment for #127


examples/podinfo/values.cue
```diff
@@ -3,6 +3,14 @@
 
 package main
 
+// Default image values
+// TODO: Implement env injection at build time
+values: image: {
+	repository: "ghcr.io/stefanprodan/podinfo"
+	digest:     "sha256:5b731f7f2662a5ebf0bed5a6d72ee5a79a4f520f199eebf6f7e927364409efd6" @timoni(env:string:image_digest)
```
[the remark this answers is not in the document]
stefanprodan: The goal for these vars are to encourage people to set both the tag and digest for all images in a module. To make people do it, it must be easy. Assuming the Timoni module is in the same repo as the app, after calling docker build-push, the tag and digest can be set into the build ENV automatically for Timoni to pick it up at `timoni mod push`.

examples/podinfo/cue.mod/pkg/timoni.sh/core/v1alpha1/image.cue
```diff
@@ -0,0 +1,19 @@
+package v1alpha1
```
errordeveloper, line 1: +1 for versioning this

errordeveloper: Would it makes sense to use `package image`, so it gets imported as that?

```
import "timoni.sh/core/v1alpha1:image"
```
WDYT?

stefanprodan: Assuming the core module will contain security context, resources and other helpers, do we want one package per type as in one package per file? In Go I would keep them under the same package.

errordeveloper: Go doesn't allow multiple packages per directory. It's just a thought. We can also say that this a niche feature of CUE that might more confusing then it is useful, but I don't have a particular opinion. I did use this feature ones in a package that contained multiple somewhat miscellaneous helpers.
I guess a reason CUE has this feature is to avoid unnecessary subdirs.

errordeveloper: From reading [the docs ](https://cuelang.org/docs/concepts/packages/#:~:text=Each%20package%20is%20identified%20by,%3A%20)%20and%20its%20package%20name.&text=If%20the%20basename%20of%20the,the%20latter%20can%20be%20omitted), I get a sense that the idea is simply so that package name doesn't have to match directory name and that there can be multiple packages in a directory.
Actually, this makes a lot of sense in contrast to Go, because there you can have a package name that differs from dir name, but that is not explicit and you actually have to know. CUE just makes this case explicit and introduces a possibility of multiple packages also.

examples/podinfo/cue.mod/pkg/timoni.sh/core/v1alpha1/image.cue
```diff
@@ -0,0 +1,19 @@
...
+
+// Image defines the schema for an OCI image reference.
+#Image: {
+	repository: string
+	tag:        string
+	digest:     string
```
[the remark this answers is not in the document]
errordeveloper: Let's add validation to digest value. 

The syntax is like this:
```
digest : string 
digest: =~ "<regex>"
```
Also, shouldn't it be optional or you prefer to avoid `!= _|_`?

PR #134, opened 2023-07-07
Add completion for instances and namespaces

jilleJr:
Adds Cobra completion on installed instances.

I've not yet added them to all commands, as I want to ensure design is good first, so I don't have that many places to edit it.

Preview:

```console
$ timoni delete <TAB>
 »» completions ««
podinfo  nginx  -- Installed in namespace "timoni-testing"

$ timoni delete -n default
 »» no matches found ««
```

Clarification: Those `»»««` headers are from Zsh I think, or some weird plugin I have installed. They're at least not from Cobra or timoni

### DIFF
--- a/cmd/timoni/list.go
+++ b/cmd/timoni/list.go
@@ -18,8 +18,9 @@ package main
 
 import (
 	"context"
-	apiv1 "github.com/stefanprodan/timoni/api/v1alpha1"
 	"io"
+
+	apiv1 "github.com/stefanprodan/timoni/api/v1alpha1"
 
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
@@ -60,22 +61,7 @@ func init() {
 }
 
 func runListCmd(cmd *cobra.Command, args []string) error {
-	sm, err := runtime.NewResourceManager(kubeconfigArgs)
-	if err != nil {
-		return err
-	}
-
-	iStorage := runtime.NewStorageManager(sm)
-
-	ctx, cancel := context.WithTimeout(context.Background(), rootArgs.timeout)
-	defer cancel()
-
-	ns := *kubeconfigArgs.Namespace
-	if listArgs.allNamespaces {
-		ns = ""
-	}
-
-	instances, err := iStorage.List(ctx, ns, listArgs.bundleName)
+	instances, err := listInstancesFromFlags()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


cmd/timoni/list.go
```diff
@@ -113,6 +100,38 @@ func runListCmd(cmd *cobra.Command, args []string) error {
...
+func completeWithInstanceList(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	instances, err := listInstancesFromFlags()
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+	completions := make([]string, len(instances))
+	for i, inst := range instances {
+		completions[i] = fmt.Sprintf("%s\tInstalled in namespace %q",
+			inst.Name, inst.Namespace)
+	}
```
stefanprodan, line 131:
```suggestion
	var completions []string
	for _, inst := range instances {
		if strings.HasPrefix(inst.Name, toComplete) {
			completions = append(completions, inst.Name)
		}
	}
```

I suggest we do completions like in kubectl.

stefanprodan:
Hey @jilleJr we also need completion for `--namespace` so we can do `timoni -n <TAB> status <TAB>`.

jilleJr:
> Hey @jilleJr we also need completion for `--namespace` so we can do `timoni -n <TAB> status <TAB>`.

Definetly. Will add, and since you don't seem to have a problem with the approach, I will add to remaining commands as well.

cmd/timoni/main.go
```diff
@@ -114,4 +116,27 @@ func addKubeConfigFlags(cmd *cobra.Command) {
 	kubeconfigArgs.Namespace = &namespace
 
 	cmd.PersistentFlags().StringVarP(kubeconfigArgs.Namespace, "namespace", "n", *kubeconfigArgs.Namespace, "The instance namespace.")
+	cmd.RegisterFlagCompletionFunc("namespace", completeNamespaceList)
+}
+
+func completeNamespaceList(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
stefanprodan, line 122: Let’s move this and the instance function to a dedicated file e.g. `completion.go`

jilleJr: Moved it to `completion_funcs.go` (a0793c6) instead of `completion.go`, to not collide with the existing `timoni completion` command

stefanprodan (approved):
LGTM

Thanks @jilleJr 🏅 

PS. Can you please squash the commits and rebase with main.

PR #137, opened 2023-07-09
Prettier and more colorized output for the apply commands

jilleJr:
I was fiddling around with the logging output. I'm used to zerolog, and know this little hack you can do by setting the `caller` field to get the text at the beginning (meant for source code line:number)

Then I started playing around with the coloring of the changes it can make, to make them stand out a bit more.

### Before

image

### After

image

### DIFF
--- a/cmd/timoni/log.go
+++ b/cmd/timoni/log.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/briandowns/spinner"
@@ -30,6 +31,7 @@ import (
 	"github.com/go-logr/zerologr"
 	gcrLog "github.com/google/go-containerregistry/pkg/logs"
 	"github.com/rs/zerolog"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	runtimeLog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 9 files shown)


stefanprodan:
@jilleJr this looks wonderful, thank you!

Could we highlight better in the --dry-run the created/modified/deleted resources vs the unchanged ones, maybe we gray out the once with no changes?

jilleJr:
> @jilleJr this looks wonderful, thank you!
> 
> Could we highlight better in the --dry-run the created/modified/deleted resources vs the unchanged ones, maybe we can gray out the once with no changes?

You mean like this?

Screenshot_20230711_161744

Or like this?

Screenshot_20230711_162143

cmd/timoni/log.go
```diff
@@ -56,6 +62,118 @@ func NewConsoleLogger(pretty bool) logr.Logger {
...
+	colorCallerPrefix = color.New(color.FgHiBlack)
+	colorBundle       = color.New(color.FgHiMagenta)
+	colorInstance     = color.New(color.FgHiMagenta)
+	colorPerAction    = map[ssa.Action]*color.Color{
+		ssa.CreatedAction:    color.New(color.FgHiGreen),
+		ssa.ConfiguredAction: color.New(color.FgHiCyan),
+		ssa.UnchangedAction:  color.New(color.FgHiBlack),
+		ssa.DeletedAction:    color.New(color.FgRed),
+		ssa.SkippedAction:    color.New(color.FgHiBlack),
+		ssa.UnknownAction:    color.New(color.FgYellow),
```
stefanprodan, line 76: These look great, thank you!

stefanprodan:
As a followup, if you have time, the same colouring could be used in `timoni status` and `timoni inspect resources`.

jilleJr:
Sure! Sounds like some cozy coding

Issue #154, opened 2023-07-24
Generate CUE definitions from Kubernetes CRD YAMLs


stefanprodan:
Currently the way module authors can add Kubernetes definitions is by importing the Go package from a controller e.g. https://timoni.sh/module/#kubernetes-crds

```sh
# download the Prometheus Operator Go APIs
go get github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1@v0.63.0

# generate the CUE types for v1 monitoring APIs
cue get go github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1
```

But not all CRD controllers have public Go packages, for example Crossplane, as @vfarcic pointed out. There are also controllers written in other languages than Go. 

Timoni could offer a command for generating Kubernetes definitions from CRDs in YAML format. 

For example:

```sh
timoni mod import crd \
-f https://github.com/cert-manager/cert-manager/releases/download/v1.13.0/cert-manager.crds.yaml
```

This would generate the definitions inside `cue.mod/gen/` as such:

```
cue.mod/
└── gen
    ├── acme.cert-manager.io
    │   ├── challenge
    │   │   └── v1
    │   │       └── types_gen.cue
    │   └── order
    │       └── v1
    │           └── types_gen.cue
    └── cert-manager.io
        ├── certificate
        │   └── v1
        │       └── types_gen.cue
        ├── certificaterequest
        │   └── v1
        │       └── types_gen.cue
        ├── clusterissuer
        │   └── v1
        │       └── types_gen.cue
        └── issuer
            └── v1
                └── types_gen.cue
```

Not sure how can this be implemented, but I guess we could extract the OpenAPI v3 schema from the CRD YAML, and use cue to parse and generate the definitions.

carlossg:
Not sure if this is what you are looking for, but this is a dirty way to get the schemas from the CRDs
https://github.com/adobe-platform/kubernetes-json-schema/blob/master/build.sh

stefanprodan:
Maybe we could use [kube-openapi](https://github.com/kubernetes/kube-openapi) together with CUE's OpenAPI https://pkg.go.dev/cuelang.org/go/encoding/openapi

sdboyer:
> Maybe we could use [kube-openapi](https://github.com/kubernetes/kube-openapi) together with CUE's OpenAPI 

[...]

PR #166, opened 2023-08-18
Add a YAML CRD to CUE IR conversion function

sdboyer:
This PR introduces a function that converts YAML bytes containing one or more CRDs into an intermediate form.

This PR is in pursuit of #154, but does not attempt any new CLI subcommands or user-facing capabilities. Everything here is implemented in `internal` because IMO the types introduced would need more refining before becoming part of an importable public surface. Rather, this is just the guts of what's required for #154. Given this conversion func and the IR it returns, i think (?) it should be a pretty straightforward path to implement such a command.

i've also made a couple critical notes as inline comments.


internal/engine/crd.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		sch := ctx.BuildFile(of)
+		cc.Schemas = append(cc.Schemas, VersionedSchema{
+			Version: ver,
+			Schema:  sch.LookupPath(defpath),
+		})
+
+		// Additional massaging of converted schemas should be done here. Lots
+		// could be done. One example: apiVersion field appears to be coming
+		// through conversion as optional.
```
sdboyer, line 149: i pretty heavily expect this to be something that we'll want to do.

For example, to handle erroneous field optionality (and probably some other issues i haven't yet noticed), the simplest thing would be to get some native CUE definitions of what all CRDs must look like committed to timoni (similar to [this](https://github.com/grafana/kindsys/blob/74f2bc8126104cd5c989a8d0e2a8539f6a8a8b2e/kindcat_custom.cue#L41-L97)).

Then, in this func, we'd unify those static definitions with the one being processed. As long as `apiVersion` is required in the static definition, it won't matter whether it comes through from openapi extraction as required or not.

internal/engine/crd_test.go
```diff
@@ -0,0 +1,264 @@
...
+	// the client submits requests to. Cannot be updated. In
+	// CamelCase. More info:
+	// https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+	kind?: string
+	metadata?: {
+		...
+	}
+
+	// BucketSpec defines the desired state of an S3 compatible bucket
+	spec?: {
```
sdboyer, line 114: Probably the most important thing to note about this approach in contrast to `cue get go` is that CRDs have all types inlined, and lack the information that would be necessary to recreate that original structure.

So for the `spec` key here, whereas `cue get go` would generate a standalone `#BucketSpec` struct in cue that corresponds to [this struct](https://github.com/fluxcd/source-controller/blob/aa370f284db70e09d8acb49d5bed4e42908b8cb3/api/v1beta1/bucket_types.go#L33), then in the overall resource schema have `spec: #BucketSpec`, here it's all just inlined into one big object.

AFAIK, there's no way around this. My experience over the past couple years tells me that dealing with references like this is generally the most subtle and confounding problem in this general area - one that really benefits from having simple, often restrictive rules.

On the plus side, for this particular case, it's not hard to imagine an approach where the CRD is the baseline that's worked from (since it's always available and most expressive), but if a go package happens to be available, the CLI command could take an additional parameter for the go package, which could then be used to reconstruct the network of types. Handwavy example:

```
timoni mod import crd \
--package flux/source \
-f https://github.com/fluxcd/source-controller/releases/download/v1.0.1/source-controller.crds.yaml \
--go-types github.com/fluxcd/source-controller/api@v1.0.1
```

(the suffix after the `api` package is an interesting question, given that go types are distributed across all those packages)

stefanprodan: Maybe we could have some conversion function that takes the raw CRD and returns:

```
#Bucket: {
	apiVersion: *"source.toolkit.fluxcd.io/v1beta2" | string
	kind:       *"Bucket" | string

	metadata: {
		name:       string
		namespace?: string
		labels?: {[string]: string}
		annotations?: {[string]: string}
	}

	spec:    #BucketSpec
	status?: #BucketStatus
}

#BucketSpec: {...}

#BucketStatus: {...}
```

CRDs without spec would be consider invalid, only status can be missing. Not sure how to wrap the raw CRD, but I think that the above format should be the `IntermediateCRD` output.

internal/engine/crd.go
stefanprodan: Here I would drop all fields except for `spec` and `status`, and inject the Kubernetes Object props:

```
apiVersion: *"source.toolkit.fluxcd.io/v1beta2" | string
kind:       *"Bucket" | string

metadata: {
    name:       string
    namespace?: string
    labels?: {[string]: string}
    annotations?: {[string]: string}
}
```

Timoni users shouldn't have to set the `apiVersion` nor `kind` in their objects, as these values are in the original CRD.

internal/engine/crd_test.go
sdboyer: Sure, spec and status can be split out generically, and we can set `apiVersion` and `kind`. I'm not sure why you're using a default rather than just a concrete value -  `kind: *"Bucket" | string` rather than just `kind: "Bucket"` - AIUI no other value would be acceptable there.

internal/engine/crd.go
sdboyer: yep, all seems good, done

internal/engine/crd_test.go
stefanprodan: Ah yes, changing the kind doesn’t make sense.

internal/engine/crd_test.go
```diff
@@ -0,0 +1,271 @@
...
+		// this source.
+		suspend?: bool
+
+		// The timeout for download operations, defaults to 60s.
+		timeout?: string | *"60s"
+		...
+	}
+
+	// BucketStatus defines the observed state of a bucket
+	status?: {
```
stefanprodan, line 184: I’m not sure if `status`, which is a read-only field, should be exported. At apply time, Timoni removes the `status`, even if it’s specified.

internal/engine/crd_test.go
```diff
@@ -0,0 +1,271 @@
...
+			...
+		}
+
+		// This flag tells the controller to suspend the reconciliation of
+		// this source.
+		suspend?: bool
+
+		// The timeout for download operations, defaults to 60s.
+		timeout?: string | *"60s"
+		...
```
stefanprodan, line 180: Can we make all definitions as closed? You shouldn’t be able to add fields to spec, undefined fields are rejected by Kubernetes API.

sdboyer: Yeah, i fixed this in my latest push. There's a TODO there around properly handling the exception cases where unspecified fields are allowed. But i think it's good enough for an initial implementation, at least. Openness/closedness is a surprisingly deep rabbit hole.

internal/engine/crd_test.go
sdboyer: Yeah, in other discussions i've had, when to include particular subelements in schemas has often been quite murky. Generally depends on the user/use case for that variant of the schema.

i'll pull it out of the overall resource def, but retain the standalone definition.

internal/engine/crd_test.go
stefanprodan: Oh yeah I totally forgot about opened fields in CRDs, they do have a special annotation: https://github.com/fluxcd/helm-controller/blob/a8c15d3700afce4309c12f37803f96ed166991b5/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml#L747

rawkode:
I wonder if we should use something like the following to fill in the metadata?

```go
gc := gocodec.New((*cue.Runtime)(ctx), nil)
meta, err := gc.ExtractType(metav1.TypeMeta{})

if err != nil {
    return nil, fmt.Errorf("error extracting metav1.TypeMeta: %w", err)
}

... crd.FillPath ...
```

Rather than building that out ourselves?

sdboyer:
> This does currently rely on a deprecated CUE API

The underlying thing that eventually calls is [`convert.GoTypeToExpr()`](https://github.com/cue-lang/cue/blob/220b44e973eff9d41f2067499f20a12b6a0a6f29/internal/core/convert/go.go#L54), which is the same thing that's called by [`context.EncodeType()`](https://pkg.go.dev/cuelang.org/go@v0.5.0/cue#Context.EncodeType). The latter isn't deprecated - perfectly fine to rely on, and i've leaned heavily on it elsewhere.

That said, i'm not sure what part of the code in the PR you're proposing be replaced with such a conversion.

rawkode:
> > This does currently rely on a deprecated CUE API
> 
> The underlying thing that eventually calls is [`convert.GoTypeToExpr()`](https://github.com/cue-lang/cue/blob/220b44e973eff9d41f2067499f20a12b6a0a6f29/internal/core/convert/go.go#L54), which is the same thing that's called by [`context.EncodeType()`](https://pkg.go.dev/cuelang.org/go@v0.5.0/cue#Context.EncodeType). The latter isn't deprecated - perfectly fine to rely on, and i've leaned heavily on it elsewhere.
> 
> That said, i'm not sure what part of the code in the PR you're proposing be replaced with such a conversion.

I noticed the Compile String with a metadata object on line 156 and was curious if we could handle that with the type?

internal/engine/crd_test.go
sdboyer: Yeah, that's the one, `x-kubernetes-preserve-unknown-fields`. From [the docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#jsonschemaprops-v1-apiextensions-k8s-io):

> `x-kubernetes-preserve-unknown-fields` stops the API server decoding step from pruning fields which are not specified in the validation schema. This affects fields recursively, but switches back to normal pruning behaviour if nested properties or additionalProperties are specified in the schema. This can either be true or undefined. False is forbidden.

IMO, this is most properly handled in `cuelang.org/go/encoding/openapi.Extract` itself. By the time we have a `cue.Value` out, the custom k8s props have been lost. It may be possible to do a hack here where we walk the AST of the original openapi in order to figure exactly which levels should be open vs. closed, but i can't currently imagine a way of doing that without a bunch of painful wheel reinvention.

I don't have a good sense of how harmful it would be to not support this property from the beginning. Would it be acceptable to ignore it for now and treat this PR as an initial pass, or would it impact most users in a way that is likely to make the feature generally unusable?

sdboyer:
> I noticed the Compile String with a metadata object on line 156

Gotcha.

Mostly, yes, we could. i settled on this because i think it's the clearest way to express what [these lines](https://github.com/stefanprodan/timoni/pull/166/files#diff-3b5613da4d31046bcef1ee31bcc54c7879802058c2699cfd36a691c067376185R153-R161) are responsible for - establish a baseline set of fields, fieldnames, field optionalities and certain concrete field values. It's possible to do in other ways (except maybe for the [dynamic optionality of `namespace`](https://github.com/stefanprodan/timoni/pull/166/files#diff-3b5613da4d31046bcef1ee31bcc54c7879802058c2699cfd36a691c067376185R158)). 

i've ended up often leaning on `CompileString()` as the best way to express a known, static foundation of CUE values that's then easy to manipulate with the Go API.

internal/engine/crd_test.go
stefanprodan: I would say it's unusable without handling `x-kubernetes-preserve-unknown-fields`, lots of CRDs have it Flux, cert-manager, etc. As a workaround, we could add an arg e,g, `closed` so users could set it to false and then all definitions are left opened, you lose the validation but at least you can generate valid objects.

sdboyer:
sorry that i letting this sit for a few days - deciding how to tackle [this](https://github.com/stefanprodan/timoni/pull/166#discussion_r1301588163) led me to decide we need a more robust test suite now. i've started on that, but then my house lost power for a few days, so...y'know :)

More code coming this week.

sdboyer:
This is now in what i think is a mergeable state. Some explanation is necessary.

High-level tl;dr, my [most recent commit](https://github.com/stefanprodan/timoni/pull/166/commits/9b682250abce7f7a35733fc420e2eb79aaa80a92):

* Removes the logic to drop ellipses
* Introduces a new table-based test where each test case narrowly exercises one aspect of field preservation, aka openness/closedness
* Solves the problem in a way that is technically correct, adequate for [the primary known use case](https://github.com/stefanprodan/timoni/pull/166#discussion_r1301588163), but is unsatisfying and merits follow-up

Now, let's talk about the problem.

The problem centers on struct closedness. That is, given this schema:

```cue
schema: {
  specifiedField: string
}
```

Is this data valid?
```json
{
    "specifiedField": "foo"
    "extraField": 42
}
```

If `schema` is open, it's valid - additional fields are fine. If it's closed, it's invalid, because `extraField` isn't explicitly specified. The concept seems simple enough, but in practice closedness is quite subtle and complicated, especially within nested structs.

For this PR, we have to consider three different that users can express an intent that an object/struct is open or closed:

* CUE
* OpenAPI
* [Structural schemas](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema), Kubernetes' variant of OpenAPI & JSON Schema

### CUE

In CUE, structs are open by default. [The spec](https://cuelang.org/docs/references/spec/#closed-structs) has a good explanation, but there are two ways of closing structs:

```cue
Open: {
  foo: string
}

// Closed when referenced elsewhere, unless referenced via embedding.
#ClosedRecursively: {
  foo: string
  nest: {
    another: string
    // extra fields not allowed in instances
  }
  // extra fields not allowed in instances
}

RootClosedNestOpen: close({
  foo: string
  nest: {
    another: string
    // extra fields permitted in instances
  }
  // extra fields not allowed in instances
})


#RootOpenNestClosed: {
  foo: string
  nest: {
    another: string
    // extra fields not allowed - ellipses affects only the nesting level
    // on which it's declared
  }
  // extra fields in instances ordinarily not allowed,
  // but ellipsis supersedes this and makes them allowed
  ...
}
```

So, in CUE:

* Regular structs are open by default.
* The only native CUE syntax for recursively changing openness/closedness is the use of [definitions](https://cuelang.org/docs/references/spec/#definitions-and-hidden-fields) - `#`-led labels.

### OpenAPI

In OpenAPI 3.0, specifying a field with only `type: "object"` defaults to open. This is comparable to the behavior of CUE's regular fields. So, these are saying the same thing:

<table>
<tr><th>CUE</th><th>OpenAPI</th><th>Valid JSON</th></tr>
<tr>
<td>

```cue
Open: {
  foo: string
}
```

</td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      properties:
        foo:
          type: string
```

</td>
<td>

```json
{
    "foo": "ohai"
    "unspecified": 42
}
```
</td>
</tr>
</table>

The above omits the `additionalProperties` field. Omitting that field is the same as saying `additionalProperties: true` or `additionalProperties: {}` (see [these docs](https://swagger.io/docs/specification/data-models/data-types/#free-form)). Thus, these are all equivalent:

<table>
<tr><th>CUE</th><th>OpenAPI</th><th>Valid JSON</th></tr>
<tr>
<td>

```cue
Open: {
  foo: string
}
```

</td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      properties:
        foo:
          type: string
```

</td>
<td>

```json
{
    "foo": "ohai"
    "unspecified": 42
}
```
</td>
</tr>
<tr>
<td></td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      additionalProperties: true
      properties:
        foo:
          type: string
```

</td>
<td></td>
</tr>
<tr>
<td></td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      additionalProperties: {}
      properties:
        foo:
          type: string
```

</td>
<td></td>
</tr>

</table>

Setting `additionalProperties: false` on an object is the equivalent of `close()` in CUE, operating only at the level of the  The following closed, nested schema definitions are also equivalent:

<table>
<tr><th>CUE</th><th>OpenAPI</th><th>Invalid JSON</th></tr>
<tr>
<td>

```cue
Closed: close({
  foo: string
  nest: close({
    another: string
  })
})
```

</td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      properties:
        foo:
          type: string
      nest:
        type: object
        properties:
          another:
            type: string
```

</td>
<td>

```json
{
    "foo": "ohai",
    "nest": {
        "another": "wat",
        "unspecified": 42
    }
}
```
</td>
</tr>
</table>

However, AFAIK neither OpenAPI nor JSON Schema has a native construct that recursively closes specified objects. So, while CUE can eliminate those `close()` expressions and just use a definition, OpenAPI has to keep the declarations at each level of the tree.

<table>
<tr><th>CUE</th><th>OpenAPI</th><th>Invalid JSON</th></tr>
<tr>
<td>

```cue
#Closed: {
  foo: string
  nest: {
    another: string
  }
}
```

</td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      additionalProperties: false
      properties:
        foo:
          type: string
      nest:
        type: object
        additionalProperties: false
        properties:
          another:
            type: string
```

</td>
<td>

```json
{
    "foo": "ohai",
    "nest": {
        "another": "wat",
        "unspecified": 42
    }
}
```
</td>
</tr>
</table>

Given an expression that can recursively close, an expression that allows superseding that recursion becomes important. Lacking recursive closedness, neither JSON Schema nor OpenAPI need to have an equivalent expression - just, don't say `additionalProperties: false`:

<table>
<tr><th>CUE</th><th>OpenAPI</th><th>Valid JSON</th></tr>
<tr>
<td>

```cue
#Closed: {
  foo: string
  nest: {
    another: string
    ...
  }
}
```

</td>
<td>

```yaml
components:
  schemas:
    Open:
      type: object
      additionalProperties: false
      properties:
        foo:
          type: string
      nest:
        type: object
        properties:
          another:
            type: string
```

</td>
<td>

```json
{
    "foo": "ohai",
    "nest": {
        "another": "wat",
        "unspecified": 42
    }
}
```
</td>
</tr>
</table>

AFAICT, CUE's OpenAPI encoder captures all of this as expected.

### Structural schema & Kubernetes

While a main goal of structural schema is centered on the defaulting to closed structs for safety, interestingly, its approach doesn't actually change anything about object _validity_ - the typical domain of schemas. Rather, the implementation relies on pruning, which is a transform that is only applied post-validation.

I can imagine reasons why this was the approach taken. It would have broken k8s installs for reasons where, the overwhelming majority of the time, the 'right' answer would just have been to remove the fields on the sender side.

However, it creates a bit of an awkward situation for our present purposes. All we can really express in CUE is the schema - not some behavior to be applied post-validation. So technically, we don't need to _do_ anything at all with this. That's the current state of this PR - do nothing with `x-kubernetes-preserve-fields`. And i think we can merge it this way.

Will follow up with a bit more context, though.

sdboyer:
Still, doing it this way is wiggling out on a technicality. Structural schema are trying to say something about closedness, even if they only achieve it via pruning. **IMO, faithfully representing the intent of CRD structural schemas in CUE requires us to upgrade this post-validation pruning behavior into actual validity rules**. What's in this PR may ultimately be enough for Timoni's purposes, in which case, great - i'll improve it when i copy this code elsewhere :) 

What's clear to me so far is that:
* Structural schema implies that all CRDs are essentially like CUE definitions - recursively closed as a baseline.
* `x-kubernetes-preserve-unknown-fields` seems similar to `...` in CUE: it indicates that the struct on which it's declared should be treated as open.

HOWEVER, [the docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/jsonschemaprops-v1-apiextensions-k8s-io) say that
> This affects fields recursively, but switches back to normal pruning behaviour if nested properties or additionalProperties are specified in the schema.

Because structural schema _also_ only allow either `properties` OR `additionalProperties` on a field (the former case being representable as a Go struct, and the latter being representable as a Go map), i think this actually may mean that `x-kubernetes-preserve-unknown-fields` may have identical semantics to `...`.

That said, some basic tests have left me confused, because a 1.27 apiserver was rejecting CRs that seemed like they should be valid according to the rules of structural schema. More investigation later, but probably not something we have to care about in Timoni.

sdboyer:
Code is now working. No caveats or compromises about the relation between CRDs and CUE: `x-kubernetes-preserve-unknown-fields` is interpreted as exactly equivalent to CUE where a struct defined within a definition (and therefore recursively closed) has `...`, resulting in it being open.

Biggest source of risk here is now captured in [this comment](https://github.com/stefanprodan/timoni/pull/166/files#diff-3b5613da4d31046bcef1ee31bcc54c7879802058c2699cfd36a691c067376185R230-R234) - jumping multiple levels up the AST tree could be buggy in the case of certain valid CRD schemas. It's not obvious to me what inputs could actually cause an issue here. The worst case scenario should be that some ellipses are absent that ought to be present.

internal/engine/crd_test.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	// This flag tells the controller to suspend the reconciliation of
+	// this source.
+	suspend?: bool
+
+	// The timeout for download operations, defaults to 60s.
+	timeout?: string | *"60s"
+	...
```
stefanprodan, line 439: @sdboyer the spec of a Bucket doesn't have `x-kubernetes-preserve-unknown-fields` so I expect for spec to be closed as demonstrated by the `root-unspecified` test.

sdboyer: hmm, good catch, I'll investigate

stefanprodan:
@sdboyer can you please run `make test` and push the changes.

sdboyer:
done!

stefanprodan (approved):
LGTM

Thanks @sdboyer for this awesome contribution 🏅 🥇 🎖️

Issue #167, opened 2023-08-22
The Bundle API should allow to read data from K8s configMaps and secrets


denist-huma:
My K8s operator knows how to create a redis instance and put a redis URL to a secret. I can supply a hostname, but the password is generated. How can I get a password from the secret? Same for CM.
For K8s apps I would use valueFrom.secretKeyRef:

```yaml
      env:
        - name: redisURL
          valueFrom:
            secretKeyRef:
              key: password
              name: redis-secret
```

For example I have a bundle:

```cue
bundle: {
	apiVersion: "v1alpha1"
	name:       "podinfo"
	instances: {
		redis: {
			module: {
				url:     "oci://ghcr.io/compose-operator-examples/modules/redisclaim"
				version: "0.1.1"
				values:
					managedConnectionSecretRef: {
						name: "redis-secret"
					}
			}
			namespace: "podinfo"
		}
		podinfo: {
		    module: url:     "oci://ghcr.io/stefanprodan/modules/podinfo"
		    module: version: "6.3.5"
		    namespace: "podinfo"
		    values: caching: {
		        enabled:  true
		        redisURL: "tcp://redis-redis:6379"
		    }
		}
	}
}
```

denist-huma:
@stefanprodan thank you for your answer! :+1: 
Let me joke a bit: if I'm to allow a developer to supply a REDIS_PASS from his/her local enviroment I guess it would be "12345" or "qwerty",  :) so it it safer to make a platform to generate it.

About your runtime API draft:
1. it is better to support mounting volumes from ConfigMap ans Secret to.
2. I think that .valueFrom.namespace "kube-system" is too permissive. A workload should read only it local namespace secret, cm

stefanprodan:
> it is better to support mounting volumes from ConfigMap ans Secret to.

I guess I got this issue wrong then, if your controller is generating the Kubernetes Secret, then mount it in the app Deployment, no Timoni runtime needed.

denist-huma:
@stefanprodan You got it right in the first time, it is me who is adding wishes one on top of another
Let's forget I've said about volumes.
I'm interested in reading an enviroment value from K8s secret or cm into the module's runtime.

PR #180, opened 2023-09-06
Introduce the Runtime API for querying the cluster for values

stefanprodan:
This PR introduces a new API called **Runtime** for fetching values from Kubernetes API that can be mapped to **Bundle** fields.  Closes: #167

## Motivation

While Timoni Bundles offer a way to specify the config values in declarative manner, not all the configuration values of an application are known ahead of time. Some values may be available at runtime, in the Kubernetes cluster where the Bundle is applied.

For example, the API token for some backend service that your app consumes is stored in a Kubernetes Secret in-cluster. When installing the application with Timoni, you may want to fetch the token from the Kubernetes Secret and use it as a config value.

## Solution

Timoni Runtime API offers a way to define values which are fetched from the Kubernetes API and mapped to fields inside a Bundle. Using CUE expressions, users can point Timoni to any Kubernetes resource including custom resources, and extract values from any field.

Currently, the Runtime API supports querying the Kubernetes API, but in the future we could add other external sources such as Secrets Managers (AWS Secrets Manager, Azure Vault, Hashicorp Vault, 1Password, etc) and HTTP APIs.

## Example

The following is an example of a Runtime definition that extracts values from the cluster.

```cue
runtime: {
	apiVersion: "v1alpha1"
	name:       "production"
	values: [
		{
			query: "k8s:v1:ConfigMap:infra:aws-info"
			for: {
				"ACCOUNT_ID": "obj.data.account_id"
				"REGION":     "obj.data.region"
			}
		},
		{
			query: "k8s:v1:Secret:infra:redis-auth"
			for: {
				"REDIS_PASS": "obj.data.password"
				"REDIS_CA":   "obj.data.\"ca.crt\""
			}
		},
		{
			query: "k8s:source.toolkit.fluxcd.io/v1:GitRepository:flux-system:cluster"
			for: {
				"GIT_REVISION": "obj.status.artifact.revision"
				"GIT_STATUS":   "[for c in obj.status.conditions if c.type == \"Ready\" {c.status}][0]"
			}
			optional: true
		},
	]
}
```

The `query` field must be in the format `k8s:<apiVersion>:<kind>:<namespace>:<name>`.
If the Kubernetes resource is global, the format is `k8s:<apiVersion>:<kind>:<name>`.
The `for` map must contain pairs of name and CUE expression.

The values from the Runtime, can be mapped to fields in a Bundle using `@timoni()` attributes.

```cue
bundle: {
	_id:   string @timoni(runtime:string:ACCOUNT_ID)
	_reg:  string @timoni(runtime:string:REGION)
	_pass: string @timoni(runtime:string:REDIS_PASS)

	apiVersion: "v1alpha1"
	name:       "podinfo"
	instances: {
		podinfo: {
			module: url: "oci://\(_id).dkr.ecr.\(_reg).amazonaws.com/modules/podinfo"
			namespace: "podinfo"
			values: caching: {
				enabled:  true
				redisURL: "tcp://:\(_pass)@redis.infra:6379"
				redisCA: string @timoni(runtime:string:REDIS_CA)
			}
		}
	}
}
```

The `@timoni(runtime:[string|number|bool]:[VAR_NAME])` CUE attribute can be placed next to a field to set its value from the runtime.

### Commands

Build the Runtime to see which values are present on the cluster:

```sh
 timoni runtime build -f runtime.cue
```

Apply the Bundle using the values from the Runtime:

```sh
timoni bundle apply -f bundle.cue --runtime runtime.cue
```

Apply the Bundle using the values from the Runtime and from local env vars:

```sh
timoni bundle apply -f bundle.cue --runtime runtime.cue --runtime-from-env
```

The  values coming from the Runtime take precedence over the Environment.

## Breaking changes

The `timoni(env:[TYPE]:[NAME])` attribute was replaced with `timoni(runtime:[TYPE]:[NAME])`. 

The local Environment is no longer used by default, users need to opt-in by using the `--runtime-from-env` flag in `timoni bundle` commands.

To migrate, replace `@timoni(env` with `@timoni(runtime` in the your bundles.


Nalum:
Does this apply only to bundles or can it also be used with a values file?

stefanprodan:
This only applies to Bundles, using plain instances and values is more for debugging purposes when you develop a module, not actual production use. I plan to rewrite the get started guide to about Bundles.

PR #192, opened 2023-09-17
Add cmd for generating CUE definitions from Kubernetes CRDs

stefanprodan:
This PR adds a new command for importing Kubernetes CRDs in Timoni's modules.

Big thanks to @sdboyer for doing all the hard work in #166 🥇 

### Command

```console
$ timoni mod import crd --help

Generate CUE definitions from Kubernetes CRDs

Usage:
  timoni mod import crd [MODULE PATH] [flags]

Examples:
  # Generate CUE definitions from a local YAML file
  timoni mod import crd -f crds.yaml

  # Generate CUE definitions from CRDs included in a remote YAML file
  timoni mod import crd -f https://github.com/fluxcd/flux2/releases/latest/download/install.yaml
```

### Example

Create a module named myapp:

```console
$ timoni mod init myapp .
4:09PM INF module initialized at myapp
```

Navigate to the module root:

```console
$ cd myapp
```

Generate CUE definitions from the cert-manager CRDs:

```console
$ timoni mod import crd -f https://github.com/cert-manager/cert-manager/releases/download/v1.13.0/cert-manager.crds.yaml
7:24PM INF generating: acme.cert-manager.io/challenge/v1
7:24PM INF generating: acme.cert-manager.io/order/v1
7:24PM INF generating: cert-manager.io/certificate/v1
7:24PM INF generating: cert-manager.io/certificaterequest/v1
7:24PM INF generating: cert-manager.io/clusterissuer/v1
7:24PM INF generating: cert-manager.io/issuer/v1
```

Inspect the generated files:

```console
$ tree cue.mod/
cue.mod/
├── gen
│   ├── acme.cert-manager.io
│   │   ├── challenge
│   │   │   └── v1
│   │   │       └── types_gen.cue
│   │   └── order
│   │       └── v1
│   │           └── types_gen.cue
│   ├── cert-manager.io
│   │   ├── certificate
│   │   │   └── v1
│   │   │       └── types_gen.cue
│   │   ├── certificaterequest
│   │   │   └── v1
│   │   │       └── types_gen.cue
│   │   ├── clusterissuer
│   │   │   └── v1
│   │   │       └── types_gen.cue
│   │   └── issuer
│   │       └── v1
│   │           └── types_gen.cue
```

Closes: #154


internal/engine/crd.go
```diff
@@ -267,12 +269,14 @@ func convertCRD(crd cue.Value) (*IntermediateCRD, error) {
...
 							Value: ast.NewIdent("#" + kname + "Spec"),
 						}
+						specref.Constraint = token.NOT
 						astutil.CopyComments(specref, x)
 						cursor.Replace(specref)
 						return false
 					case "status":
-						statusf = new(ast.Field)
-						*statusf = *x
+						//TODO: decide if status should be included
```
sdboyer, line 277: yeah this felt wonky while i was working on it

cmd/timoni/mod_import_crd_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+
+	"github.com/mattn/go-shellwords"
+	. "github.com/onsi/gomega"
+)
+
+func TestImportCrd(t *testing.T) {
```
sdboyer, line 30: this looks like it effectively covers the same functionality as the modified tests from `internal/engine`, right? and with much cleaner golden files than that crappy string

cmd/timoni/mod_import_crd.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+
+	crds, err := engine.YamlCRDToCueIR(cuectx, b)
+	if err != nil {
+		return err
+	}
+
+	for _, crd := range crds {
```
sdboyer, line 93: awesome! a loop like this is exactly what i was hoping you were aiming to do, and what i tailored the CUE IR func towards

cmd/timoni/mod_import_crd_test.go
stefanprodan: Oh yeah, this gives you a readable output which is easy to understand since it uses `git diff`. You see spot the differences and if needed you can also regenerate the golden files with a locally built binary.

cmd/timoni/mod_import_crd.go
stefanprodan: I moved the loop to the engine now in af2b9a8 to remove the `_Def` and give consumers key value pairs, contain a unique identifier in the format `<group>/<kind>/<version>` and the contents of the CUE definition.

I also realised that the CUE IR function needs valid Kubernetes CRDs, so in 72d0f06 I'm parsing the multi-doc YAML and I'm extracting only valid CRDs. This way the engine function always receives what's expected, a multi-doc with CRDs and nothing else.

internal/engine/crd.go
stefanprodan: Yeah I decided to remove status for now, Timoni strips the status at apply-time same as kubectl, so having it there just inflates the files with no real benefit.

PR #240, opened 2023-11-10
docs: update secret doc to use FIFO instead of temporary file

b4nst:
This PR updates documentation to show usage of the `sops exec-file` command, which avoid potential leaks of secrets by using a FIFO. Secrets are only in-memory decrypted.


stefanprodan (approved):
LGTM

Thanks @b4nst 🏅

Issue #251, opened 2023-11-21
Documentation on using Cue and Kubernetes


jmgilman:
First of all, thanks for this interesting project, we're currently evaluating using it internally at our company. That being said, the documentation is great for getting started with Timoni, but for users new to CUE + Kubernetes, there doesn't seem to be much for getting started. In other words, I could read both the Timoni and CUE docs and still have very few examples of how the Timoni docs came up with some of the given examples. 

Are there resources out there that I am missing?

stefanprodan:
There is nothing specific in CUE for Kubernetes, Timoni comes with Kubernetes schemas that you can use to do resource templating, which work like any other CUE value. Please see https://timoni.sh/module/ and let me know what things you would like to see in there specific to Kubernetes. I suggest starting with `timoni mod init app .` and trying to add a new resource like in the docs.

jmgilman:
```
❯ timoni mod init myapp .
4:02PM INF module initialized at myapp
❯ cd myapp                                                                      
❯ cue cmd -t debug -t name=test -t namespace=test -t mv=1.0.0 -t kv=1.28.0 build
import failed: expected '}', found 'EOF':
    ./timoni.cue:11:2
    ./templates/config.cue:68:3
❯ cue version                                                                   
cue version 0.5.0

go version go1.20.8
      -buildmode exe
       -compiler gc
       -trimpath true
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
```

Unfortunately, I can't get linting to work, but I don't see anything immediately wrong with the specified file.

jmgilman:
Just a few initial comments from a complete beginner:

The result of `mod init` is _very_ large. I understand the value of a deployable example, but there's so much wiring going on that it took a while for me to understand how things were being compiled together. In addition to that, because things are so tightly wired, removing some of the example resources ended up breaking things in multiple places. For example, what if I don't want a test job? I was surprised at how much I had to change when I deleted the job file. 

Perhaps an additional flag to `mod init` that adds the complete example would be more appropriate. Something like the following seems to be the minimal amount needed:

- cue.mod
- templates
   - templates/config.cue
- debug_*.cue
- timoni.cue
- timoni.ignore
- values.cue

As it stands, for me to "get started" with my first project, I had to spend quite a bit of time unwiring the starting template.

stefanprodan:
> I realized there's a newer version of cue which fixed the issue (the NixOS package is behind).

[...]

PR #261, opened 2023-11-30
Document the module config required fields

stefanprodan:
Changes:
- Explain which `#Config` fields are used by Timoni to inject the name, namespace, module version and cluster version.
- Explain which changes are acceptable for Timoni's entry point.
- Move the Kubernetes min version constraint to `#Config`.

Ref: #251


docs/module.md
```diff
@@ -75,28 +75,28 @@ timoni: {
...
 	// Pass the generated Kubernetes resources
 	// to Timoni's multi-step apply.
-	apply: all: [ for obj in instance.objects {obj}]
+	apply: app: [ for obj in instance.objects {obj}]
 }
 ```
 
-### Kubernetes min version
+The entry point is largely considered read-only, the only field that can be modified
+if the `apply` instruction.
```
jmgilman, line 85: if -> is

examples/minimal/timoni.cue
jmgilman, line 1: Just a note on this:

```cue
if instance.config.test.enabled {
	apply: test: [ for obj in instance.tests {obj}]
}
```

Even though this is a conditional, the module author doesn't really get a chance to "optionally" write tests. The reason being:

1. The conditional references `config.test` which makes it a required field
2. If you leave `config.test` and just don't implement tests, the consumer of the package is left wondering what this `test` option does (and would probably be disappointed it does nothing). 

So the options are you either implement tests or unwire all of the logic around tests (which IIRC was in a few more places). This is a very minor nitpick, so it's probably fine as is, but this was just my experience coming into it (perhaps we'll adopt the testing with jobs strategy, but it's not something we've historically practiced in this way).

stefanprodan: Oh yeah that's the issue with the current blueprint, which turned out it's not quite minimal. Will work on that next, planning to have more flavours for the init command and also allow custom blueprints.

PR #270, opened 2023-12-06
Add `#MetaComponent` generator to Timoni's CUE schemas

stefanprodan:
This PR adds generators for Kubernetes object metadata that can be used to set a prefix to the `metadata.name` and injects the `app.kubernetes.io/component` label.

The `#MetaComponent` exposes a `#LabelSelector` property that can be used for `selector.matchLabels` and `template.metadata.labels` to set the `app.kubernetes.io/name` and `app.kubernetes.io/component` labels.

Example:
```cue
#Frontend: appsv1.#Deployment & {
	_config: #Config
	_meta: timoniv1.#MetaComponent & {
		#Meta:      _config.metadata
		#Component: "frontend"
	}
	apiVersion: "apps/v1"
	kind:       "Deployment"
	metadata:   _meta
	spec: appsv1.#DeploymentSpec & {
		replicas: _config.replicas
		selector: matchLabels: _meta.#LabelSelector
		template: {
			metadata: {
				labels: _meta.#LabelSelector
			}
			spec: corev1.#PodSpec & {...}
		}
	}
}
```

The `#MetaClusterComponent` can be used for non-namespaced objects such as ClusterRole.

PR #271, opened 2023-12-06
Add `#MatchLabelsComponent` generator to Timoni's CUE schemas

Nalum:
This PR adds a generator for Kubernetes object selectors that will inject the `app.kubernetes.io/component` label to the selector. The example below makes use of the `#MetaComponent` introduce in #270.

Example:

```cue
#Deployment: appsv1.#Deployment & {
        _config:    #Config
        _component: "test"

        apiVersion: "apps/v1"
        kind:       "Deployment"

        _metadata: timoniv1.#MetaComponent & {
                #Meta:      _config.metadata
                #Component: _component
        }

        metadata: _metadata

        spec: appsv1.#DeploymentSpec & {
                replicas: _config.replicaCount
                selector: timoniv1.#MatchLabelsComponent & {
                        #SelectorLabels: _config.selector.labels
                        #Component:      _component
                }

                template: {
                        metadata: labels: _metadata.labels

                        if _metadata.annotations != _|_ {
                                metadata: annotations: _metadata.annotations
                        }
                }
        }
}
```

### DIFF
--- a/schemas/timoni.sh/core/v1alpha1/selector.cue
+++ b/schemas/timoni.sh/core/v1alpha1/selector.cue
@@ -3,6 +3,8 @@
 
 package v1alpha1
 
+import "strings"
+
 // Selector defines the schema for Kubernetes Pod label selector used in Deployments, Services, Jobs, etc.
 #Selector: {
 	// Name must be unique within a namespace. Is required when creating resources.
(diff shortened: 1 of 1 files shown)


stefanprodan:
@Nalum please see https://github.com/stefanprodan/timoni/pull/270 that exposes the selector directly from Metadata without the need for an extra helper. Also in your example above, you set the template labels to the metadata.labels, but these should be set to the selector value.

PR #292, opened 2023-12-20
Add `--values` arg to `timoni mod vet` cmd

### LINKED ISSUE #288: [Proposal] Replace --debug with --file for vet
Currently the `vet` command only accepts a `--debug` boolean argument to be able to vet the module against `debug_values.cue` file.
I find this file quite confusing, and often removes it all together with the `debug_tool.cue`. I think it would be useful for the `vet` command to take a `--values` (or `--value` if we want to keep it simple) argument, and vet against those(this) values file(s).

It would help us testing complex module where there's mutually exclusive configuration value (cloud provider for example).

What do you think @stefanprodan? If ok with you I can push a PR

b4nst:
resolves #288 but keeps the `debug` arg.

### DIFF
--- a/cmd/timoni/mod_vet.go
+++ b/cmd/timoni/mod_vet.go
@@ -50,10 +50,11 @@ var vetModCmd = &cobra.Command{
 }
 
 type vetModFlags struct {
-	path  string
-	pkg   flags.Package
-	debug bool
-	name  string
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @b4nst

Issue #273, opened 2023-12-08
Nested packages in templates is difficult


jmgilman:
The issue is a bit hard to explain but it is mostly relevant to large packages which produce a lot of resources. Given the default Timoni structure:

- timoni.cue
- templates/
   - config.cue
   - file1.cue
   - file2.cue

Let's say that the templates folder has grown very large (>10 files at the root). The `templates` package is also starting to become overcrowded. So, we introduce a new subpackage:

- timoni.cue
- templates/
   - config.cue
   - file1.cue
   - file2.cue
   - subpkg/
      - file3.cue

Now, here we run into a problem:

1. Each of the files contains structures that rely on `#Config` being passed in. This is normal and is used in all of the Timoni examples.
2. The `#Instance` structure in `config.cue` also relies on these individual files, as they declare the resources being generated by the module. For example, `#Instance` might use `subpkg.#Deployment`.
3. So, for example, `config.cue` imports `timoni.sh/my_module/subpkg` to access the resource structures and `file3.cue` imports `timoni.sh/my_module/templates` to access the `#Config` structure. This creates a cyclic dependency.

The only reason this appears to be a problem is because `#Instance` is included with `#Config` in the same package. If, for example, we moved `#Instance` to another package, it could have a dependency on `#Config` and all of the resources in the `subpkg` package without creating a cyclic dependency. 

It's very possible I'm missing something obvious here, or there is another way to solve this issue, but my limited CUE knowledge isn't really showing me anything. From what I'm seeing, it's impossible to easily introduce additional packages containing resource structures without running into a cyclic dependency every time.

jmgilman:
```mermaid
flowchart LR
    A[config.cue]-- #Instance depends on #Service--->C[subpkg/file3.cue #Service]
    C-- #Service depends on #Config-->A
```

The issue arises when we introduce a sub-package:

```mermaid
flowchart LR
    A[config.cue]-- import 'timoni.sh/mod/templates/subpkg' --->C[subpkg/file3.cue #Service]
    C-- import 'timoni.sh/mod/templates' --->A
```

If we, for example, move `#Instance` to the root-level, we resolve the circular dependency:

```mermaid
flowchart LR
    A[instance.cue]-- import 'timoni.sh/mod/templates/subpkg' --->C[subpkg/file3.cue #Service]
    A[instance.cue]-- import 'timoni.sh/mod/templates' --->B[config.cue #Config]
    C-- import 'timoni.sh/mod/templates' --->B
```

[...]

PR #302, opened 2023-12-30
Change the `config` type to definition

stefanprodan:
This PR renames the `_config` field used in templates to `#config`.

The redis example module was refactor to showcase how to use nested packages for templates.

Fix: #273


jmgilman:
LGTM.

PR #307, opened 2024-01-04
docs: specifies where the instance inventory is stored

### LINKED ISSUE #306: Document how inventory is managed
When creating a new instance using `timoni apply` the documentation mentions the following:

```
Creates or updates the instance inventory with the last applied resources IDs.
```

I cannot find any additional information on how this is being managed. Where is the instance inventory stored in the cluster? In one recent case, I managed to create a resource that Timoni was unable to delete without an error. This caused the instance to never truly delete and it wasn't apparent to me where Timoni was looking to find this inventory information so I could manually purge it.

jmgilman:
Fixes #306. This is minimal documentation, but it should unblock someone trying to troubleshoot a broken inventory by hinting at where it's stored.


stefanprodan (approved):
LGTM

Thanks @jmgilman

Issue #309, opened 2024-01-06
Timoni fails to load kubeconfig when the KUBECONFIG environment variable contains multiple config paths


algobot76:
OS: MacOS 14.2.1 (M3)

timoni version:

api: timoni.sh/v1alpha1
client: 0.18.0
cue: 0.7.0

![image](https://github.com/stefanprodan/timoni/assets/20016835/d46326dc-6508-4d6f-a82f-2e1496f007ab)

I have an idea about how to support multi-path KUBECONFIG. I may submit a PR for this issue later next week.

stefanprodan:
You could merge the two into a single file, then pass that to Timoni:

```
kubectl config view --flatten > merged.conf
timoni --kubeconfig=merged.conf
```

algobot76:
It works, but I think it could be better if it cloud select the corresponding config based on the current context.

Issue #300, opened 2023-12-29
ERR loading kubeconfig failed: stat ~/.kube/config:: no such file or directory


JCzz:
os: macos arm64 m1

installed with brew
timoni -v
timoni version 0.18.0

timoni -n test apply podinfo oci://ghcr.io/stefanprodan/modules/podinfo --version latest
2:30PM INF i:podinfo > pulling oci://ghcr.io/stefanprodan/modules/podinfo:latest
2:30PM INF i:podinfo > using module timoni.sh/podinfo version 6.5.4
2:30PM ERR loading kubeconfig failed: stat /Users/christian/.kube/config:: no such file or directory

The file is there: "~/.kube/config"

stefanprodan:
Do you have the `KUBECONFIG` environment variable set by any chance? Can you please try:
```
timoni -n test apply podinfo oci://ghcr.io/stefanprodan/modules/podinfo \
--kubeconfig ~/.kube/config
```

JCzz:
Thanks, yes that worked. Strange now that kubectl is working on the same cluster.

Nive work and looking forward to follow you work.

stefanprodan:
It may be related to #309 Do you have multiple config files in the `KUBECONFIG` env var?

PR #310, opened 2024-01-07
Find the correct kubeconfig based on thee current context

algobot76:
Fix: #309
Fix: #300

### DIFF
--- a/cmd/timoni/main.go
+++ b/cmd/timoni/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"os"
 	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/fatih/color"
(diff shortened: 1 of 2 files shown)


cmd/timoni/main.go
```diff
@@ -152,3 +155,33 @@ func addKubeConfigFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(kubeconfigArgs.Namespace, "namespace", "n", *kubeconfigArgs.Namespace, "The the namespace scope for the operation.")
 	cmd.RegisterFlagCompletionFunc("namespace", completeNamespaceList)
 }
+
+func getCurrentKubeconfigPath() string {
+	defaultPath := path.Join(homedir.HomeDir(), ".kube", "config")
```
stefanprodan, line 160: I think we can set `defaultPath := ""`. The `cli-runtime` package that we use in Timoni has the home dir logic when the client is initialised with an empty kubeconfig path, so we don't need to import `k8s.io/client-go/util/homedir`.

algobot76: Done, please take a look.

stefanprodan (approved):
LGTM

Thanks @algobot76 🏅

PR #326, opened 2024-01-26
doc/fix cue command parameter order

fberube:
fix the parameter order
from: 
`cue import /path/to/manifests.yaml  -o -f templates/manifests.cue -p templates -l 'strings.ToLower(kind)' -l 'metadata.name'`
to 
`cue import /path/to/manifests.yaml -f -o templates/manifests.cue -p templates -l 'strings.ToLower(kind)' -l 'metadata.name'`


stefanprodan (approved):
LGTM

Thanks @fberube

PR #334, opened 2024-02-02
Module Markdown Optional Fields

nalum:
Based on feedback in https://github.com/cue-lang/cue/discussions/2783 this PR updates the Markdown document generation to pull the optional fields into the table data. This is done via the `cue.Value.Fields()` iterator.


internal/engine/module_builder.go
```diff
@@ -315,52 +315,91 @@ func (b *ModuleBuilder) GetConfigDoc(value cue.Value) ([][]string, error) {
...
+		v := fields.Value()
+		fmt.Println(fields.Label(), v.IncompleteKind(), v.IsConcrete())
+
+		// We are chekcing if the field is a struct and not optional and is concrete before we iterate through it
+		// this allows for definition of default values as full structs without generating output for each
+		// field in the struct where it doesn't make sense e.g.
+		//
+		// - annotations?: {[string]: string}
+		// - affinity: corev1.Affinity | *{nodeAffinity: requiredDuringSchedulingIgnoredDuringExecution: nodeSelectorTerms: [...]}
+		if v.IncompleteKind() == cue.StructKind && !fields.IsOptional() && v.IsConcrete() {
```
Nalum, line 348: Do we add another convention, like `+nodoc`, that would force rendering of `metadata: labels:`?

With this check for struct and optional we loose `metadata: labels:` from the output.

stefanprodan: Having `+nodoc` sounds great!

Nalum: Sorry, I'll expand on this.

So this already supports `+nodoc`, the issue I've hit which brought this up is that `metadata: labels:` has data in it and so the generated output doesn't contain `metadata: labels:` if the data in it has `+nodoc` set.

What I'm thinking to do is add another `+` label that would force adding a field to the output instead of diving into it if it's an object. Something like `+value`, not sure of anything already existing that would make sense.

Not at my machine at the moment but will put together an example to show you when I can.

stefanprodan: Ah I get now what you are saying, assuming the parent struct has `+nodoc` there are cases when we want to include some of the struct fields. In this case we could use `+doc` that would take precedence over the parent exclusion.

Nalum: Actually the other way around, we want to have the labels field documented but not the labels in it. By having those labels marked `+nodoc` the label field is not documented.

Nalum: So for example we have the following in the timoni schema:

```cue
// Metadata defines the schema for Kubernetes object metadata.
#Metadata: {
        ...
        // Map of string keys and values that can be used to organize and categorize (scope and select) objects.
        // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels
        labels: #Labels

        // Standard Kubernetes labels: app name, version and managed-by.
        labels: {
                "\(#StdLabelName)":      name
                "\(#StdLabelVersion)":   #Version
                "\(#StdLabelManagedBy)": "timoni"
        }

        // LabelSelector selects Pods based on the app.kubernetes.io/name label.
        #LabelSelector: #Labels & {
                "\(#StdLabelName)": name
        }
}
```

```markdown
| `metadata: labels: app.kubernetes.io/name` | `testing` |
| `metadata: labels: app.kubernetes.io/version` | `vX.Y.Z` |
| `metadata: labels: app.kubernetes.io/managed-by` | `timoni` |
```

We then update this with `+nodoc` and as follows:

```cue
// Metadata defines the schema for Kubernetes object metadata.
#Metadata: {
        ...
        // Map of string keys and values that can be used to organize and categorize (scope and select) objects.
        // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels
        labels: #Labels

        // Standard Kubernetes labels: app name, version and managed-by.
        labels: {
                // +nodoc
                "\(#StdLabelName)":      name
                // +nodoc
                "\(#StdLabelVersion)":   #Version
                // +nodoc
                "\(#StdLabelManagedBy)": "timoni"
        }

        // LabelSelector selects Pods based on the app.kubernetes.io/name label.
        #LabelSelector: #Labels & {
                "\(#StdLabelName)": name
        }
}
```

The labels will not be part of the output now but neither will `metadata: labels:` because it was entered into as a structure field. I'll take another look at this and see if I can sort out this issue without needing another keyword. But if not another keyword would be the approach I'm thinking of at the moment. Which would remove the need to have `+nodoc` on each label as well e.g.

```cue
// Metadata defines the schema for Kubernetes object metadata.
#Metadata: {
        ...
        // Map of string keys and values that can be used to organize and categorize (scope and select) objects.
        // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels
        // +doc
        labels: #Labels

        // Standard Kubernetes labels: app name, version and managed-by.
        labels: {
                "\(#StdLabelName)":      name
                "\(#StdLabelVersion)":   #Version
                "\(#StdLabelManagedBy)": "timoni"
        }

        // LabelSelector selects Pods based on the app.kubernetes.io/name label.
        #LabelSelector: #Labels & {
                "\(#StdLabelName)": name
        }
}
```

```markdown
| `metadata: labels: ` | `{"app.kubernetes.io/name": "testing",...}` |
```

Nalum: I guess I could check the output of the function to see if there are any rows returned and if not then put that field in place, could avoid needing a new `+` tag though would need `+nodoc` on all fields in a structure. Will test this tomorrow.

Nalum: @stefanprodan I've updated this so that it will now add a row for any struct that has `+nodoc` on all child fields. Not sure if this is the best approach as it means it needs to be structured in a certain way e.g.

```cue
// Metadata defines the schema for Kubernetes object metadata.
#Metadata: {
        ...
        // Map of string keys and values that can be used to organize and categorize (scope and select) objects.
        // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels
        labels: #Labels

        // Standard Kubernetes labels: app name, version and managed-by.
        // +nodoc
        labels: "\(#StdLabelName)": name
}
```

This will not do what we want as `+nodoc` is being applied to `labels` (I believe), it would need to be structured like this:

```cue
// Metadata defines the schema for Kubernetes object metadata.
#Metadata: {
        ...
        // Map of string keys and values that can be used to organize and categorize (scope and select) objects.
        // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels
        labels: #Labels

        // Standard Kubernetes labels: app name, version and managed-by.
        labels: {
                // +nodoc
                "\(#StdLabelName)":      name
        }
}
```

Nalum:
@stefanprodan with some feedback from the cue team I've gotten this so it now will print the definition of the field, see the README for the blueprints starter module. I want to refine the module so that name and namespace are not output, but wanted to bring it to your attention.

Nalum:
I'm still not getting what I want from the labels and annotations though :/

Nalum:
I have pulled and rebased the latest main, this behaves in the same way. I think the changes here are worth merging, I'll spend some time digging into cue releases since the initial work on this and see if I can get the result I initially wanted for annotations and labels.

stefanprodan:
Hey @Nalum with this we remove the `default` column from the readme, would be possible to preserve it or we need to merge type with defaults? If so, I think we need to rename the type column to something else.

Nalum:
I removed it because the type information have use the type and the default in one result, I had not found a way to get both separately at the time, maybe there is a way now, I'll dig into that too.

Do you want to hold off on merging this for now so?

Nalum:
Ah I think the goal was to show the cue type information as is. For example if we just had the type and default this would not show you the possible values:
```markdown
| `test: startupAPICheck: service: type:` | `*"ClusterIP" \| "NodePort" \| "LoadBalancer" \| "ExternalName"` |
```
You would have this instead:
```markdown
| `test: startupAPICheck: service: type:` | `string` | `ClusterIP` |
```

Nalum:
I guess this adds a requirement to understand cue syntax :thinking:

PR #335, opened 2024-02-02
Use shorter syntax for string keys in CUE schemas

errordeveloper:
This avoid uncessary interpolation, which makes for longer syntax, but it is generally just uncessary.


errordeveloper:
@stefanprodan how do I update `examples/*/cue.mod/gen`?

stefanprodan:
Copy over the files from schemas dir to examples and most importantly to blueprints.

stefanprodan (approved):
LGTM

Thanks @errordeveloper 

PS. I will sync the APIs with the blueprints before next release.

PR #339, opened 2024-02-07
chore: add proto plugin

b4nst:
This PR adds a plugin file to be able to install `timoni` in a [proto](https://moonrepo.dev/docs/proto) toolchain. It follows [toml plugin](https://moonrepo.dev/docs/proto/toml-plugin) format.

This is the file I'm currently using locally, but I thought it would be nice to add it here directly so we can reference timoni [here](https://moonrepo.dev/docs/proto/tools).


stefanprodan (commented):
Can you please add proto to the install page as a tab.

b4nst:
Sure! done.

stefanprodan (approved):
LGTM

Thanks @b4nst

Issue #268, opened 2023-12-05
Implement `timoni mod build`


jmgilman:
Currently, it appears the only way to publish a Timoni module in OCI format is to use `timoni mod push`. This is unfortunate, as it limits what an end-user can do with the OCI artifact prior to pushing it to a registry. This is especially true in existing CI systems, where the desire may be to integrate Timoni as another step (i.e. to produce container images, but not necessarily publish them until later).

In this sense, I believe something like `timoni mod build` would be a helpful intermediate step that performs the same operations as `timoni mod publish` but without pushing the image to a registry.

jmgilman:
@stefanprodan FYI I will see about attempting this if the use case sounds good to you. If you have any pointers feel free to share them. It will probably take me a bit to read through the code base.

stefanprodan:
@jmgilman you could add a new function to the OCI package, same code as in push, but after this line dump the image to disk instead of pushing it: https://github.com/stefanprodan/timoni/blob/da003ac3714b941f37933b76604adcc155781902/internal/oci/push_module.go#L108

rp-thomas:
@stefanprodan have another question on this. Would it make sense to be able to reference the my-module.tgz in a bundle file?

Usecase: I'm currently working on a bunch of different modules for a bigger application and want to bundle these in a bundle.cue.
If I want to test the bundle creation e.g. `timoni bundle build` or apply it in a local cluster, I have to publish each module every time to the registry before I can test the whole bundle.

stefanprodan:
>  Inside each bundle file the path is relative to that bundle file.

But a bundle can be composed of many files at different paths e.g.:

```
timoni bundle apply -f ./some/path/to/bundle.cue -f /some/other/path/bundle.extas.cue
```

After we unify the bundle definition which would be the root for the relative paths?

b4nst:
Nah what I was saying is that if we setup the CUE build properly, I think we can still find reference to the original file even after unification.

So everything in `./some/path/to/bundle.cue` should be relative to `./some/path/to/bundle.cue` and everything in `/some/other/path/bundle.extras.cue` would be relative to `/some/other/path/bundle.extras.cue`

b4nst:
Okay I just confirmed ☝️. I made a quick `bundle debug` command, which is basically a copy pasta of `bundle build` but as soon as it get to `v, err := bm.Build()` I shortened it to:

```go
[...]

PR #348, opened 2024-02-19
Add support for file based module in bundle

b4nst:
Hey folks! This PR brings local module support to Bundles. Since it's quite a refactor, I'd really appreciate your input before I go any further. Hence why it's lacking tests, documentation and such.

> [!IMPORTANT]
> Heads up, this PR involves some hefty changes to the module fetcher.

While putting together the changes we discussed privately and in #268, I realized it's going to bulk up the current `Fetcher` module quite a bit. So, I've gone ahead and created a new `fetcher` package within the `engine`. This should make things more manageable and set the stage for adding new fetcher types down the road (like an HTTPS fetcher). What do you think, @stefanprodan?

At the moment, the `Local` fetcher only handles directories (think `timoni apply`). I didn't want to cram too much into one PR, so I've held off on adding `.tar.gz` file support for now. I'll tackle that in a separate PR.

I've given it a test spin with a modified version of podinfo, which I've linked here for you to try out (it uses the local redis). If this approach aligns with our goals, it could be a winner.

Before I mark this PR as ready, I'll beef up the testing (especially in the new fetcher package), flesh out the documentation, and tidy up the `BundleBuilder` for easier maintenance. I'd love to hear your thoughts on these next steps!


internal/engine/fetcher/local.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (f *Local) Fetch() (*apiv1.ModuleReference, error) {
+	if fs, err := os.Stat(f.src); err != nil || !fs.IsDir() {
+		return nil, fmt.Errorf("module not found at path %s", f.src)
+	}
+
+	modFile := path.Join(f.src, "cue.mod", "module.cue")
+	timoniFile := path.Join(f.src, "timoni.cue")
+	valuesFile := path.Join(f.src, "values.cue")
```
stefanprodan, line 60: Aren't these declared in ` NewLocal` as requiredFiles?

examples/bundles/podinfo.cue
```diff
@@ -8,11 +8,7 @@ bundle: {
 	name:       "podinfo"
 	instances: {
 		redis: {
-			module: {
-				url:     "oci://ghcr.io/stefanprodan/modules/redis"
-				version: "7.2.2"
-				digest:  "sha256:22ab5f4f0158f02b353c168f21a54901a45b2a56865fcc470e5492120deb4111"
-			}
+			module: url: "file://../redis"
```
stefanprodan, line 11: I would undo this change and properly document this feature in https://timoni.sh/bundle/#working-with-bundles under a section called "Using local modules".

stefanprodan (commented):
This looks great @b4nst just a couple of comments from me

internal/engine/fetcher/local.go
b4nst: Yes indeed ahaha. I declared them exactly for that purpose 🤦 . I'll cleanup that.

examples/bundles/podinfo.cue
b4nst: Sounds fair! Wilco

internal/testutils/registry.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		t.Fatalf("failed to create docker registry: %s", err)
+	}
+
+	go dockerRegistry.ListenAndServe()
+
+	return address
+}
+
+// func SetupTestRegistry(t *testing.T) string {
```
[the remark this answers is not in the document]
b4nst: Oops thanks for catching that, done

stefanprodan (approved):
LGTM

Thanks @b4nst this is an awesome feature 🏅 

PS. Going to add e2e tests in CI for local bundles in a followup PR.

Issue #349, opened 2024-02-21
Typo in config, reference to #ObjectReference


ametad:
This reference misses a `#`:

https://github.com/stefanprodan/timoni/blob/cba7a5df7d23ff9897f68d8591fc2bfcecbfa7e2/examples/minimal/templates/config.cue#L80

Corrected:

```cue
imagePullSecrets?: [...timoniv1.#ObjectReference]
```

PR #350, opened 2024-02-21
Correct reference to `timoniv1.#ObjectReference`

ametad:
Closes https://github.com/stefanprodan/timoni/issues/349


stefanprodan (approved):
LGTM

Thanks @ametad

PR #360, opened 2024-03-04
Add docs for referencing local modules in bundles

stefanprodan:
Add example on how to create bundles using local module paths.


GeorgeMac:
I was about to come ask if this was possible or request it. Love it 😍

Issue #375, opened 2024-03-28
Metadata is missing finalizers field


jmgilman:
The [metadata schema](https://github.com/stefanprodan/timoni/blob/main/schemas/timoni.sh/core/v1alpha1/metadata.cue#L22) currently does not support adding the `finalizers` field (which is valid). I ran into this when generating `Application` resources for Argo CD:

```cue
#Application: appv1alpha1.#Application & {
	#config:    config.#Config
	apiVersion: "argoproj.io/v1alpha1"
	kind:       "Application"
	metadata: #config.metadata & {
		if #config.cascadeDelete {
			finalizers: [
				"resources-finalizer.argocd.argoproj.io",
			]
		}
	}
	// ....
}
```

If you want to enable cascading deletes in Argo, [you must declare a finalizer upfront](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/#about-the-deletion-finalizer). While this might be a bit of an edge case, I don't see a reason for not supporting it in the definition:

```cue
#Metadata: {
    // ....
    finalizers?: [...string]
}
```

PR #379, opened 2024-04-05
Add finalizers to `#Metadata` definition

stefanprodan:
Add optional `finalizers` field to the `#Metadata` definition part of the Timoni's CUE schemas.

Fix: #375


jmgilman:
LGTM

PR #401, opened 2024-06-12
Update `cuelang.org/go` to v0.9.1

myitcv:
This PR only exists to verify a change that will be released as part of cuelang.org/go@v0.9.1. Specifically, it enables a compatibility mode where legacy modules (which do not have a language.version field) can be consumed.

For more information see https://cuelang.org/issue/3219


stefanprodan:
@myitcv this looks great! Big thanks for the fix.

myitcv:
Great, thanks. Just to re-emphasise: please DON'T submit this PR. We'll tag v0.9.1 tomorrow (assuming we don't run into any other issues that need fixing). At which point we can re-do the PR to use the actual v0.9.1 release.

errordeveloper:
@stefanprodan do you think new modules should start adding 'language: version: v0.9.1' as well?

stefanprodan:
@errordeveloper yes the blueprints used by `timoni mod init` will have the language version after next release.

myitcv:
> 'language: version: v0.9.1' as well?

Let me come back to you to confirm whether you should set `language: version: "v0.9.0"` or `language: version: "v0.9.1"` (assuming that is you upgrade to v0.9.1 when it is released).

stefanprodan:
@myitcv I guess the version in module.cue is like for Go, and sets the minimum version supported? For example, a module with `language: version: "v0.9.0"` will work with CUE v0.10.0?

myitcv:
> @myitcv I guess the version in module.cue is like for Go, and sets the minimum version supported? For example, a module with `language: version: "v0.9.0"` will work with CUE v0.10.0?

Correct, I just wanted to double check our plan with respect what `cue mod init` will do when we release v0.9.1 (because that's effectively the command that you are emulating in Timoni). 

To confirm, in CUE v0.9.1, `cue mod init` will include the following line:

```
language: version: "v0.9.0"
```

So Timoni should do the same. Please feel free to mark this hard-coding as a "TODO" on your side, because we are also going to expose the following via public API: 

https://github.com/cue-lang/cue/blob/859f1f6daadaaf200ab10702f93af03fd97329b8/internal/cueversion/version.go#L20

At which point you can remove the hard-coding.

stefanprodan (approved):
LGTM

Thanks @myitcv 🥇

PR #404, opened 2024-06-24
Post-refactoring cleanup

errordeveloper:
A cleanup a few minor bugs from #366.

### DIFF
--- a/internal/reconciler/reconciler.go
+++ b/internal/reconciler/reconciler.go
@@ -113,7 +113,7 @@ func (r *Reconciler) Init(ctx context.Context, builder *engine.ModuleBuilder, bu
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


internal/reconciler/reconciler.go
```diff
@@ -174,12 +174,6 @@ func (r *Reconciler) doWaitForTermination(_ context.Context, _ logr.Logger, cs *
 }
 
 func (r *Reconciler) ApplyAllSets(ctx context.Context, log logr.Logger, withChangeSet withChangeSetFunc) error {
-	if !r.instanceExists {
```
errordeveloper, line 177: There is exactly the same invocation in `ApplyInstance`, which is actually visible in this diff right above.

stefanprodan (approved):
LGTM

Thanks @errordeveloper

PR #429, opened 2024-10-02
[bundle] Fix local module refs for multi-cluster

### LINKED ISSUE #364: Local module reference fails when applying multi-clusters
When applying a bundle containing a local module reference with a multi cluster runtime, only the first cluster passes. The second one always fails with this kind of error:

```trace
11:55AM ERR module not found at path /var/folders/vl/j712bqm55gd754rr5tdm8bxm0000gn/T/timoni3063563812/gcp-eu/examples/redis
```

A simple test can be run in this repository by creating this bundle and runtime at the root of the repo:

```cue
// bundle.cue
bundle: {
	apiVersion: "v1alpha1"
	name:       "debug"
	instances: {
		redis: {
			module: url: "file://./examples/redis/"
			namespace: "default"
			values: maxmemory: 256
		}
	}
}
```

```cue
// runtime.cue
runtime: {
	apiVersion: "v1alpha1"
	name:       "production"
	clusters: {
		"first": {
			kubeContext: "local"
			group:       "test1"
		}
		"second": {
			kubeContext: "local"
			group:       "test2"
		}
	}
}
```

```shell
timoni bundle apply -f bundle.cue --runtime runtime.cue --diff
```

huguesalary:
This bug is caused by the relationship between the following components:
* The `func (b *BundleBuilder) InitWorkspace(workspace string,
  runtimeValues map[string]string) error` function in
  `internal/engine/bundle_builder.go`
* The `BundleBuilder.files` field
* The `BundleBuilder.mapSourceToOrigin` field
* The `func (b *BundleBuilder) getInstanceUrl(v cue.Value) string`
  function in `internal/engine/bundle_builder.go`

When called, the `InitWorkspace` goes through the list of files stored
in `b.files`, parses them and writes the result in new files (let's call
them "result files") located under the path indicated by `workspace`. In
parallel, `b.mapSourceToOrigin` maps a "result file" path to the
original file the result is based on.

At the end of the function, the original list `b.files` is replaced with
the list of "result files".

The next invocation of `InitWorkspace` then uses `b.files` again, but
this time it contains the previously computed "result files" paths.

So, for a given:
* `b.files == ["/path/to/a/bundle.cue"]`

calling `InitWorkspace("/tmp/workspace1/")`, will yield:
* `b.files == ["/tmp/workspace1/bundle.cue"]`
* `b.mapSourceToOrigin["/tmp/workspace1/bundle.cue"]="/path/to/a/bundle.cue"`

Then calling `InitWorkspace("/tmp/workspace2/")`, will yield:
* `b.files == ["/tmp/workspace2/bundle.cue"]`
* `b.mapSourceToOrigin["/tmp/workspace2/bundle.cue"]="/tmp/workspace1/bundle.cue"`

The `getInstanceUrl` relies on the `mapSourceToOrigin` being accurate to
propely resolve relative `file://` paths.

When `getInstanceUrl()` is called for the first bundle, it will propely
resolve `./examples/redis/` to `/path/to/a/examples/redis/`.

When `getInstanceUrl()` is called for the second bundle, it will erroneously
resolve `./examples/redis/` to `/tmp/workspace1/examples/redis/`.

This commit fixes this.

However it is worth noting that using a relative path for a `file://`
URI feels strange. Should it be allowed in the first place?

Fix #364

### DIFF
--- a/internal/engine/bundle_builder.go
+++ b/internal/engine/bundle_builder.go
@@ -37,7 +37,7 @@ import (
 type BundleBuilder struct {
 	ctx               *cue.Context
 	files             []string
-	workspaceFiles    map[string][]string
+	workspacesFiles   map[string][]string
 	mapSourceToOrigin map[string]string
 	injector          *RuntimeInjector
 }
@@ -64,7 +64,7 @@ func NewBundleBuilder(ctx *cue.Context, files []string) *BundleBuilder {
 	b := &BundleBuilder{
 		ctx:               ctx,
 		files:             files,
-		workspaceFiles:    make(map[string][]string),
+		workspacesFiles:   make(map[string][]string),
 		mapSourceToOrigin: make(map[string]string, len(files)),
 		injector:          NewRuntimeInjector(ctx),
 	}
@@ -118,13 +118,13 @@ func (b *BundleBuilder) InitWorkspace(workspace string, runtimeValues map[string
 		files = append(files, dstFile)
 	}
 
-	schemaFile := filepath.Join(workspace, fmt.Sprintf("%v.schema.cue", len(b.workspaceFiles[workspace])+1))
+	schemaFile := filepath.Join(workspace, fmt.Sprintf("%v.schema.cue", len(b.workspacesFiles[workspace])+1))
 	files = append(files, schemaFile)
 	if err := os.WriteFile(schemaFile, []byte(apiv1.BundleSchema), os.ModePerm); err != nil {
 		return err
 	}
 
-	b.workspaceFiles[workspace] = files
+	b.workspacesFiles[workspace] = files
 	return nil
 }
 
@@ -137,7 +137,7 @@ func (b *BundleBuilder) Build(workspace string) (cue.Value, error) {
 		DataFiles: true,
 	}
 
-	ix := load.Instances(b.workspaceFiles[workspace], cfg)
+	ix := load.Instances(b.workspacesFiles[workspace], cfg)
 	if len(ix) == 0 {
 		return value, fmt.Errorf("no instances found")
 	}
--- a/internal/engine/bundle_builder.go
+++ b/internal/engine/bundle_builder.go
@@ -37,6 +37,7 @@ import (
 type BundleBuilder struct {
 	ctx               *cue.Context
 	files             []string
+	workspaceFiles    map[string][]string
 	mapSourceToOrigin map[string]string
 	injector          *RuntimeInjector
 }
@@ -63,6 +64,7 @@ func NewBundleBuilder(ctx *cue.Context, files []string) *BundleBuilder {
 	b := &BundleBuilder{
 		ctx:               ctx,
 		files:             files,
+		workspaceFiles:    make(map[string][]string),
 		mapSourceToOrigin: make(map[string]string, len(files)),
 		injector:          NewRuntimeInjector(ctx),
 	}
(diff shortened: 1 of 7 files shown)


huguesalary:
This fix doesn't work if `-f -` is used instead of an actual path to a file.

internal/engine/bundle_builder.go
```diff
@@ -157,16 +156,16 @@ func (b *BundleBuilder) Build() (cue.Value, error) {
...
-			source = origin
+		dir, err := os.Getwd()
+		if err != nil {
+			return "", err
 		}
-		url = apiv1.LocalPrefix + filepath.Clean(filepath.Join(filepath.Dir(source), path))
+		url = apiv1.LocalPrefix + filepath.Clean(filepath.Join(dir, path))
 	}
-	return url
+	return url, nil
```
huguesalary, line 168: I'm not convinced this is the right way to fix this. It feels weird to call `os.Getwd()` from inside the library. 

It seems to me the Timoni cmd package should somehow provide all the information the library needs. So, maybe `BundleBuilder` should have a new field like `cwd` pointing to the current `cwd`, instead of trying to figure it out itself.

stefanprodan:
> However it is worth noting that using a relative path for a file:// URI feels strange. Should it be allowed in the first place?

This is how many tools work, for example Docker Compose, so I don't find it strange but actually useful.

internal/engine/bundle_builder.go
```diff
@@ -160,11 +161,7 @@ func (b *BundleBuilder) Build() (cue.Value, error) {
 func (b *BundleBuilder) getInstanceUrl(v cue.Value) string {
 	url, _ := v.String()
 	if path := strings.TrimPrefix(url, apiv1.LocalPrefix); IsFileUrl(url) && !filepath.IsAbs(path) {
-		source := v.Pos().Filename()
-		if origin, ok := b.mapSourceToOrigin[source]; ok {
-			source = origin
-		}
-		url = apiv1.LocalPrefix + filepath.Clean(filepath.Join(filepath.Dir(source), path))
+		url = apiv1.LocalPrefix + filepath.Clean(filepath.Join(b.cwd, path))
```
stefanprodan, line 164: We can't use CWD here, the relative path is to the bundle file (what ` v.Pos().Filename()` does) instead of the working dir of the CLI. You can see the e2e tests failing for: 

```
timoni bundle build -f hack/tests/nginx_bundle.cue --runtime-from-env

ERR module not found at path /home/runner/work/blueprints/starter
```
```

huguesalary: > We can't use CWD here, the relative path is to the bundle file (what  v.Pos().Filename() does) instead of the working dir of the CLI.

Ha ok, I understand the logic better now, thanks.

This raises the issue that when the bundle file comes from Stdin (`-f -`), the `v.Pos().Filename()` becomes the temporary directory of the file Stdin was written to. Which makes the `getInstanceUrl` function fail again, since the path is now wrong.

I'm also unclear what will happen when someone specifies both a bundle via Stdin and a file (`-f- -f bundle.cue`).

I'll look into that.

stefanprodan: I think we can error out if the bundle comes from stdin and contains local refs.

huguesalary: Sounds good.

I removed the `cwd` code and added 4 new tests:
* bundle apply with file from stdin (`-f-`) and `module: url: file://./relative/path/to/module`
* bundle apply with file from stdin (`-f-`) and `module: url: file:///absolute/path/to/module`
* bundle apply with file from concrete file (`-f /path/to/bundle.cue`) and `module: url: file://./relative/path/to/module`
* bundle apply with file from concrete file (`-f /path/to/bundle.cue`) and `module: url: file:///absolute/path/to/module`

How's that?

stefanprodan (approved):
LGTM

Thanks @huguesalary 🥇

PR #434, opened 2024-10-16
Honor the log related cli parameters

tired-old-man:
When `timoni` cli is used in automation, color formatting presents challenges in parsing the output.
Unfortunately, log parameters are ignored.

Currently when logger is initialized, the (true, true) pair as passed as parameters.
This results in following command line parameters being ignored:
- `--log-color`
- `--log-pretty`

This simple fix resolve logger initialization.

Fix: #436


stefanprodan (approved):
LGTM

Thanks @tired-old-man 🥇

PR #443, opened 2024-11-15
Fix CLI autocompletion on ZSH when installed with Homebrew

natemccurdy:
## Summary

Prior to this, when installing Timoni using Homebrew on a Mac with ZSH,
CLI autocompletion would not work.

After this change, CLI autocompletion works after installation or
reinstallation via Homebrew.

The problem is that the completion function that Homebrew wrote to
`/opt/homebrew/share/zsh/site-functions/` did not start with an
underscore, `_`.

By convention, completion function names (and the files containing them)
should start with an underscore followed by the name of the command they
provide completions for. When ZSH's completion system is initialized
(usually through the compinit function), it automatically looks for
files in the fpath directories that start with an underscore.

## Before this change

Autocompletion on MacOS with ZSH did not work:
broken

## After this change

Autocompletion on MacOS with ZSH works. To test this, I just renamed `/opt/homebrew/share/zsh/site-functions/timoni` to `/opt/homebrew/share/zsh/site-functions/_timoni` then opened a new shell.
fixed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -69,7 +69,7 @@ brews:
       bash_output = Utils.safe_popen_read(bin/"timoni", "completion", "bash")
       (bash_completion/"timoni").write bash_output
       zsh_output = Utils.safe_popen_read(bin/"timoni", "completion", "zsh")
-      (zsh_completion/"timoni").write zsh_output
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @natemccurdy

PR #452, opened 2024-12-03
Add --digest flag to `timoni apply` and `timoni build`

### LINKED ISSUE #450: `timoni apply` could accept a `--digest` flag and error out if the digest of the pulled module is different
Just like in bundles, `timoni apply` could accept a `--digest` flag and error out if the digest of the pulled module is different.

If the `--version` is not provided, `timoni apply` should pull through the digest URL and figure out the version from the image annotation.

matheuscscp:
Fixes #450

### DIFF
--- a/cmd/timoni/apply.go
+++ b/cmd/timoni/apply.go
@@ -102,6 +102,7 @@ type applyFlags struct {
 	module             string
 	version            flags.Version
 	pkg                flags.Package
+	digest             flags.Digest
 	valuesFiles        []string
 	dryrun             bool
 	diff               bool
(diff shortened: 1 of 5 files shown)


cmd/timoni/apply.go
```diff
@@ -116,6 +117,7 @@ var applyArgs applyFlags
 func init() {
 	applyCmd.Flags().VarP(&applyArgs.version, applyArgs.version.Type(), applyArgs.version.Shorthand(), applyArgs.version.Description())
 	applyCmd.Flags().VarP(&applyArgs.pkg, applyArgs.pkg.Type(), applyArgs.pkg.Shorthand(), applyArgs.pkg.Description())
+	applyCmd.Flags().VarP(&applyArgs.digest, applyArgs.digest.Type(), applyArgs.digest.Shorthand(), applyArgs.digest.Description())
```
stefanprodan, line 120: Please add the same flag to `timoni build`

matheuscscp: added, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #424, opened 2024-09-10
Fix  CRD import issue with `oneOf`, `not`, and `anyOf`

errordeveloper:
CUE now has resolved the CRD import issue with `oneOf`, `not`, and `anyOf` (#336) as well required fields. This PR integrates latest revision and updates tests.

Fix: #336

PR #458, opened 2024-12-14
Bumps CUE to version 0.11.0

### LINKED ISSUE #457: Update CUE API
The current version of the [CUE Go API](https://github.com/stefanprodan/timoni/blob/53b4f415520546a95925b68c37511a13e36872ed/go.mod#L6) appears to be quite outdated. This is also confirmed via the CLI:

```
❯ timoni version
api: timoni.sh/v1alpha1
client: 0.22.1
cue: 0.9.2
```

I am attempting to use symlinks in my `cue.mod/pkg` directory, but support for this was added after `0.9.2`. I recommend bumping to the latest stable version: `0.11.0`.

jmgilman:
Fixes #457. 

There was a lot of work done on the JSON Schema decoder in v0.11.0. The primary difference, in the case of Timoni, is that generated CRDs now properly use `!` where the spec marks a field as required. As a result, I had to update a few tests to reflect this change.

Other than that, all other tests passed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup CUE
         uses: cue-lang/setup-cue@a93fa358375740cd8b0078f76355512b9208acb1 # main
         with:
-          version: v0.9.2
+          version: v0.11.0
       - name: Run tests
         run: make test
       - name: Run linter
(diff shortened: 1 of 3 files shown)


stefanprodan:
Hey @jmgilman the CUE update is included with #424 The major change in v0.11 is that the CRD conversion to CUE schemas have been fixed.

stefanprodan:
@jmgilman could you please rebase

jmgilman:
@stefanprodan Done.

stefanprodan (approved):
LGTM

Thanks @jmgilman

PR #475, opened 2025-02-08
Add TIMONI_CACHE_DIR variable as another method to set the cache location

folliehiyuki:
This is useful in CI/CD environment, where one can set the variable globally to a custom location inside the project directory. The cache directory then can be cached as an artifact to be reused by later jobs.

This approach is cleaner than appending `--cache-dir` flag into every timoni command within the CI config file.


stefanprodan (approved):
LGTM

Thanks @folliehiyuki

PR #485, opened 2025-02-19
docs: Remove reference to non-existent flag --prune

edeustua:
Remove reference to non existent `timoni bundle apply` flag `--prune`.


stefanprodan (approved):
LGTM

Thanks @edeustua

PR #490, opened 2025-03-03
fix: use `list.Concat()` to concatenate lists

ccmtaylor:
cue v0.11 removed support for using the + and * operators on lists, causing the `cue cmd` flags listed in the doc strings to fail. Update the examples to use `list.Concat()` instead.


stefanprodan (approved):
LGTM

Thanks @ccmtaylor

PR #547, opened 2026-06-19
Fix capabilities element type

zhshishkin:
Trying timoni for my project I noticed that I can not change capabilities because of type mismatch. This PR if applied fixes its type in both example and blueprint


stefanprodan (approved):
LGTM

Thanks @zhshishkin

PR #558, opened 2026-07-22
fix: bump defaultKubeVersion

till:
I tried to add a test, but that requires parsing `go.mod`. Not sure if you want to, it's a bit ugly. But anyway. Here it is!


stefanprodan (approved):
LGTM

PR #564, opened 2026-07-31
fix(oci): Preflight registry tags

reneleonhardt:
## What

- Validate OCI Distribution tag syntax in artifact references and tag values.
- Reject every invalid `--tag` before input processing or registry I/O.
- Document the requirement.

## Why

The go-containerregistry parser accepts leading `.` and `-` tags that OCI Distribution rejects. Preflight prevents invalid artifacts from being built or published.

## Reproduction

```shell
! ./bin/timoni artifact push oci://registry.example.com/org/artifact -f /does/not/exist -t 1.0.0 -t .invalid
# invalid OCI tag ".invalid"
```

## Verification

- `go test ./internal/oci -run 'Test(ValidateTag|ParseArtifactURLTags|TagArtifactRejectsInvalidTagBeforeURL)'`
- `go test ./cmd/timoni -run 'Test(TagArtifactRejectsInvalidTagsBeforeRegistry|PushArtifactRejectsInvalidTagsBeforeInput)'` (requires envtest)

### DIFF
--- a/cmd/timoni/artifact_push.go
+++ b/cmd/timoni/artifact_push.go
@@ -92,6 +92,7 @@ func init() {
 	artifactCmd.AddCommand(pushArtifactCmd)
 }
 
+// pushArtifactCmdRun validates tags, builds, and pushes an OCI artifact.
 func pushArtifactCmdRun(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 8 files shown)


stefanprodan (approved):
LGTM

PR #565, opened 2026-07-31
fix(mod): Centralize version validation

reneleonhardt:
## What

- Validate module versions through one shared check in `mod build` and `mod push`.
- Reject empty, malformed, and build-metadata versions before local output or publishing.
- Correct the OIDC issuer regexp flag name in the bundle distribution example.

## Why

A module version is also its OCI tag. Both commands reject build metadata before side effects.

## Reproduction

```shell
! ./bin/timoni mod build testdata/module -v 1.0.0+demo -o ./tmp/module.oci.tar
# version build metadata is not supported
! ./bin/timoni mod push /does/not/exist oci://localhost:5000/module -v 1.0.0+demo
# version build metadata is not supported
```

## Verification

- `go test ./internal/flags -run TestValidateModuleVersion`
- `go test ./cmd/timoni -run 'Test_(BuildMod|PushMod)RejectsBuildMetadataVersion'` (requires envtest)
- `rg --fixed-strings -- '--certificate-oidc-issuer-regexp' docs/bundle-distribution.md`
- `! rg --fixed-strings -- '--certificate-oidc-issuer-regex=' docs/bundle-distribution.md`

### DIFF
--- a/cmd/timoni/mod_build.go
+++ b/cmd/timoni/mod_build.go
@@ -20,13 +20,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/spf13/cobra"
 
 	apiv1 "github.com/stefanprodan/timoni/api/v1alpha1"
 	"github.com/stefanprodan/timoni/internal/engine"
+	"github.com/stefanprodan/timoni/internal/flags"
 	"github.com/stefanprodan/timoni/internal/logger"
 	"github.com/stefanprodan/timoni/internal/oci"
 )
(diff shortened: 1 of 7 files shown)


stefanprodan (approved):
LGTM

stefanprodan:
@reneleonhardt I plan to do a release today with all these fixes and improvements. Do you want to delay the release? Do you have more in pipeline or its good to go?

reneleonhardt:
Don't wait for me 😅
No, I didn't know what's especially important to you at the moment, like a roadmap.
Those were the easiest fixes my Codex could find.

Is there some kind of blocker/feature I should give a shot at? I want to use Timoni in production after all those opinionated helm problems.
Or should I increase coverage significantly? codemap does a pretty good job tracing all code paths / hubs... after I created dozens of patches to make it work inside a sandbox 😅

reneleonhardt:
The only real correctness issues Codex could find are at the moment:

## Next candidates

1. `fix(crd): Preserve unknown fields in array items`
   `convertCRD` tracks preservation paths while traversing object properties, but not schema array items. An `x-kubernetes-preserve-unknown-fields` object in an array can lose its open CUE ellipsis. Add a failing importer fixture, then traverse items with the matching list selector. Scope: `internal/engine/importer.go`, `internal/engine/importer_test.go`. Verify: `go test ./internal/engine -run TestConvertCRD`.

2. `fix(oci): Write module cache layers atomically`
   `PullModule` uses `Stat` plus `os.Create` for a shared digest cache. Concurrent processes can read a partial layer; a failed reader can remove a layer another process is writing. Write to a temporary file, close it, then publish atomically. Scope: `internal/oci/pull_module.go`, `internal/oci/module_test.go`. This needs a short design review for concurrent publishers and Windows replacement semantics.

3. `fix(fetcher): Allow uncached OCI fetching`
   `PullModule` supports an empty cache directory with a temporary directory, but `fetcher.OCI.Fetch` first calls `MkdirAll` on it. Pass the empty value through and add a focused fetcher test. Scope: `internal/engine/fetcher/oci.go`, `internal/engine/fetcher/oci_test.go`.

stefanprodan:
The first one reads like a major bug, not sure if there are actual CRDs doing this. I maintain a catalog of CNCF projects schemas here: https://github.com/controlplaneio-fluxcd/schema-catalog you could ask Codex to look for a CRD that reproduces the bug.

All 3 finds seem worth fixing!

reneleonhardt:
I started them, thanks for the schemas!

PR #567, opened 2026-07-31
fix(oci): Validate module cache layers

reneleonhardt:
## What

- Verify cached and downloaded layers against their manifest digest.
- Publish complete layers atomically and repair corrupt cache entries.

## Why

Concurrent fetches share digest-named cache files. Readers could observe partial or corrupt layers, while extraction cleanup could remove a file another process owns.

## Reproduction

`go test ./internal/oci -run 'Test(ModuleOperations|WriteCachedLayer)$' -count=1`

Before this patch, the corrupt-cache regression fails with `gzip: invalid header`. The atomic-write test also verifies that concurrent writers publish only complete, valid content.

## Verification

`go test ./internal/oci -count=1`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/mattn/go-shellwords v1.0.13
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/onsi/gomega v1.42.1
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

PR #568, opened 2026-07-31
fix(fetcher): Allow uncached OCI fetching

reneleonhardt:
## What

Pass an empty cache directory through to the OCI module puller so it can use an ephemeral cache.

## Why

Disabling caching leaves the cache directory empty, but the OCI fetcher called `MkdirAll("")` and failed before pulling the module.

## Reproduction

`go test ./internal/engine/fetcher -run 'TestOCIFetch/without_cache' -count=1`

Expected stderr before this patch: `mkdir : no such file or directory`.

## Verification

`go test ./internal/engine/fetcher -count=1`


stefanprodan (approved):
LGTM

PR #566, opened 2026-07-31
fix(engine): Preserve unknown CRD array fields

reneleonhardt:
## What

- Traverse OpenAPI array items when collecting preservation paths.
- Preserve unknown fields only when the Kubernetes extension is `true`.

## Why

Some CRDs, including OpenKruise SidecarSet, preserve unknown fields on array items. Timoni generated closed CUE objects for those items.

## Reproduction

```shell
timoni mod init module .
timoni mod vendor crd ./module -f https://raw.githubusercontent.com/openkruise/kruise/828b8ca4c2c0176b76b2cd43284cb74010e10ad1/config/crd/bases/apps.kruise.io_sidecarsets.yaml
awk '/^[[:space:]]*containers\?: \[\.\.\.\{/ {show=1} show {print; text=$0; sub(/\/\/.*/, "", text); opens=gsub(/\{/, "", text); closes=gsub(/\}/, "", text); depth+=opens-closes; if (depth==0) exit}' ./module/cue.mod/gen/apps.kruise.io/sidecarset/v1alpha1/types_gen.cue
# main closes the array item; this branch emits `...`.
```

## Verification

- `go test ./internal/engine -count=1`
- `make test`

Developed with carefully directed, manually reviewed AI assistance.

Co-Authored-By: GPT-5.6 Sol <codex@openai.com>

PR #569, opened 2026-07-31
fix(engine): Preserve unknown CRD map value fields

stefanprodan:
Followup: #566 

Keep map value objects open for CRDs such as OpenFeature FeatureFlag and Kubewarden AdmissionPolicyGroup, which set x-kubernetes-preserve-unknown-fields on schemas nested under additionalProperties.

### DIFF
--- a/internal/engine/importer.go
+++ b/internal/engine/importer.go
@@ -254,6 +254,12 @@ func convertCRD(crd cue.Value) (*IntermediateCRD, error) {
 					return err
 				}
 			}
+			if sch.AdditionalProperties.Schema != nil {
+				nextPath := append(path[:len(path):len(path)], cue.AnyString)
+				if err := walkfn(nextPath, sch.AdditionalProperties.Schema.Value); err != nil {
+					return err
+				}
+			}
 
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


stefanprodan:
@reneleonhardt can you please review this PR? Thanks

reneleonhardt:
The fix looks correct. One small test would lock the important map/list selector boundary, since `AnyString` and `AnyIndex` both stringify as `[_]`:

```diff
diff --git a/internal/engine/importer_test.go b/internal/engine/importer_test.go
--- a/internal/engine/importer_test.go
+++ b/internal/engine/importer_test.go
@@ -498,0 +499,8 @@
+
+func TestPreservePathMatchesSelectorTypes(t *testing.T) {
+	g := NewWithT(t)
+	suffix := cue.Str("variants")
+
+	g.Expect(preservePathMatches(cue.MakePath(cue.AnyString, suffix), cue.MakePath(cue.Index(0), suffix))).To(BeFalse())
+	g.Expect(preservePathMatches(cue.MakePath(cue.AnyIndex, suffix), cue.MakePath(cue.Str("key"), suffix))).To(BeFalse())
+}
```

Otherwise, the production change and real OpenFeature and Kubewarden paths look good.

PR #570, opened 2026-07-31
fix(engine): Return CUE lookup errors

reneleonhardt:
## What

Return the selected CUE value's lookup error instead of the compiled root's nil error.

## Why

A missing expression reports a generic downstream error instead of the lookup failure.

## Reproduction

```shell
printf 'other: true\n' | timoni build app ./examples/redis -f -
# main: `undefined value`
# here: `field not found: values`
```

## Verification

`go test ./internal/engine -run TestExtractValueFromBytesLookupError -count=1`

### DIFF
--- a/internal/engine/utils_test.go
+++ b/internal/engine/utils_test.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"testing"
 
+	"cuelang.org/go/cue/cuecontext"
 	. "github.com/onsi/gomega"
 )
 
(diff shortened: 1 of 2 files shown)


internal/engine/utils_test.go
```diff
@@ -69,3 +70,10 @@ func TestIsFileUrl(t *testing.T) {
 	g.Expect(IsFileUrl("file://afile.txt")).To(BeTrueBecause("file:// is a file URL"))
 	g.Expect(IsFileUrl("oci://foo/bar")).To(BeFalseBecause("oci:// is not a file URL"))
 }
+
+func TestExtractValueFromBytesLookupError(t *testing.T) {
+	g := NewWithT(t)
+
+	_, err := ExtractValueFromBytes(cuecontext.New(), []byte("other: true"), "values")
+	g.Expect(err).To(HaveOccurred())
```
stefanprodan:
```suggestion
	g.Expect(err).To(MatchError(ContainSubstring("not found"))))
```

reneleonhardt:
If you like, `Squash and merge` my branches if you only want one squash commit.
I only let Codex reshape my branches again and again to make it easier for you to review them 😉

https://docs.github.com/en/pull-requests/reference/pull-request-merges#squash-and-merge-your-commits

internal/engine/utils_test.go
reneleonhardt: Can I revert that? It' doesn't compile. Or can you add a fixed suggestion?

PR #582, opened 2026-08-01
fix(oci): Bound Cosign subprocesses

reneleonhardt:
## What

Drain Cosign output concurrently, preserve all subprocess errors, and honor caller cancellation and `--timeout`.

## Why

Reading stdout before stderr can deadlock when Cosign fills the stderr pipe. Descendants could keep those pipes open after cancellation, while a nonzero exit discarded output-processing errors.

## Reproduction

```shell
docker run --rm -d --name timoni-registry -p 5001:5000 registry:2
tmp="$(mktemp -d)"
mkdir "$tmp/bin" "$tmp/artifact"
printf artifact > "$tmp/artifact/data"
printf '%s\n' '#!/bin/sh' 'i=0' 'while [ "$i" -lt 20000 ]; do' '  echo stderr >&2' '  i=$((i + 1))' 'done' > "$tmp/bin/cosign"
chmod +x "$tmp/bin/cosign"
PATH="$tmp/bin:$PATH" artifact push oci://localhost:5001/demo -t 1.0.0 -f "$tmp/artifact" --sign=cosign --registry-insecure --creds=anonymous
# main: hangs after publishing while the fake Cosign fills stderr
# here: exits successfully
docker stop timoni-registry
```

## Verification

`go test ./internal/oci -run '^TestProcessCosignIO' -count=1`

`go test ./cmd/timoni -run '^TestRootCommandPreservesContextCancellation$' -count=1`

Developed with carefully directed, manually reviewed AI assistance.

Co-Authored-By: GPT-5.6 Sol <codex@openai.com>

PR #585, opened 2026-08-01
fix(oci): Preflight and bound Cosign subprocesses

reneleonhardt:
## Summary

- Resolve the configured Cosign executable before registry or filesystem side effects.
- Bound and concurrently drain Cosign subprocess output while preserving cancellation and process errors.
- Supersedes #582, which contained the first commit in this PR.

## Verification

- `go test ./cmd/timoni/... ./internal/oci/...`

Signed-off-by: Rene Leonhardt <65483435+reneleonhardt@users.noreply.github.com>

### DIFF
--- a/cmd/timoni/artifact_pull.go
+++ b/cmd/timoni/artifact_pull.go
@@ -108,6 +108,8 @@ func pullArtifactCmdRun(cmd *cobra.Command, args []string) error {
 	ociURL := args[0]
 
 	log := LoggerFrom(cmd.Context())
+	ctx, cancel := context.WithTimeout(cmd.Context(), rootArgs.timeout)
+	defer cancel()
 
 	if _, err := oci.ParseArtifactURL(ociURL); err != nil {
 		return err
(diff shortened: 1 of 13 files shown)


stefanprodan (commented):
`cosign verify` emits a JSON array of all signatures on a single line which can be bigger than the 64KB bufio.

I would  raise the cap so the limit is a log-truncation guard rather than a failure mode, and stop promoting logging faults on the success path:

```go
  // in scanCosignOutput
  scanner := bufio.NewScanner(output)
  scanner.Buffer(make([]byte, 0, bufio.MaxScanTokenSize), 1024*1024)

  // in processCosignIO
  if err := errors.Join(stdoutErr, stderrErr); err != nil {
      log.Error(err, "cosign output could not be fully logged")
  }
  return nil
```

reneleonhardt:
Consolidated review path: #582 is closed because this PR includes its exact commit plus the related executable-preflight commit. I am evaluating the 64 KB scanner/logging finding here as a follow-up revision.

reneleonhardt:
Implemented the 64 KB finding in a2e6199. The scanner accepts Cosign output payload lines through 1 MiB (with the required one-byte terminator allowance); scan faults are drained and logged as `cosign output could not be fully logged` rather than failing a successful Cosign process. Process and cancellation failures still return. Verified with `make test`.

stefanprodan (approved):
LGTM

Thanks @reneleonhardt 🏅

reneleonhardt:
You're welcome, thanks for maintaining a top-notch helm alternative and being open to contributions!

Next 5 flights inbound, brace for impact...

PR #586, opened 2026-08-01
fix(oci): Bound layer extraction

reneleonhardt:
## What

Extract OCI layers with Flux's 100 MiB limit and skip archived symlinks.

## Why

Layer extraction disabled Flux's size limit and failed when an archive contained symlinks.

## Reproduction

```shell
tmp_dir="$(mktemp -d)"
dd if=/dev/zero of="$tmp_dir/large" bs=1048576 count=101
artifact push oci://localhost:5000/example -t 1.0.0 -f "$tmp_dir" --registry-insecure
artifact pull oci://localhost:5000/example:1.0.0 -o "$tmp_dir/output" --registry-insecure
# main: extracts the layer
# here: stderr contains 'bigger than max archive size'
```

## Verification

```shell
go test ./internal/oci -run TestExtractLayer -count=1
```

### DIFF
--- a/internal/oci/extract.go
+++ b/internal/oci/extract.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2026 Stefan Prodan
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


stefanprodan:
Hmm why not enable symlink skip in `fluxcd/pkg/tar` and set `DefaultMaxUntarSize`? Hand rolling this over battle tested libs feels wrong to me, unless you found a CVE in Flux? If so, please follow the disclosure policy we've published here: https://fluxcd.io/security/

reneleonhardt:
Rewritten to use `fluxcd/pkg/tar` with `DefaultMaxUntarSize` and `WithSkipSymlinks`; the custom extractor is gone. Added focused coverage for both policies, and the full test suite passes.

stefanprodan (approved):
LGTM

PR #587, opened 2026-08-01
fix(cli): Support registry credential files

reneleonhardt:
## What

Allow `--creds @<path>` to read registry credentials from a file of up to 64 KiB. Remove one terminal LF or CRLF, and use `@@` for a literal leading `@`.

## Why

Inline credentials expose registry passwords in shell history and process arguments.

## Reproduction

```shell
artifact list oci://127.0.0.1:1/example --creds @/tmp/missing-registry-creds
# main: stderr contains 'listing tags failed'
# here: stderr contains 'reading credentials file'
```

## Verification

```shell
go test ./internal/flags -run TestCredentialsSet -count=1
```

### DIFF
--- a/cmd/timoni/artifact_push.go
+++ b/cmd/timoni/artifact_push.go
@@ -42,8 +42,8 @@ the ignore rules will be used to exclude files from the artifact.`,
   timoni artifact push oci://docker.io/org/app -t latest -f .
 
  # Push a dir contents to GitHub Container Registry using a GitHub token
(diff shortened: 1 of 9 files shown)


stefanprodan:
I don't find this feature useful to anyone. If your auth is on disk, then it should be in a Docker config, not in file that contains `user:pass`.

reneleonhardt:
Understood. Timoni already uses Docker’s standard credential resolution when `--creds` is omitted, so I’ll close this in favor of Docker config and credential helpers.

PR #588, opened 2026-08-01
fix(cli): Bound and secure CRD downloads

reneleonhardt:
## What

Limit remote CRD manifests to 16 MiB after HTTP content decoding, reject redirects to plain HTTP, and cancel downloads with the command context.

## Why

Remote manifests were read into memory without a size bound, could follow HTTPS redirects to plain HTTP, and ignored caller cancellation until the command timeout.

## Reproduction

```shell
mod vendor crd ./module -f https://server.example/crds-over-16mib.yaml
# main: downloads the complete response before parsing
# here: stderr contains 'response exceeds the 16777216-byte limit'
```

## Verification

```shell
go test ./cmd/timoni -run 'TestReadRemoteCRDManifest' -count=1
```

### DIFF
--- a/cmd/timoni/mod_vendor_crd.go
+++ b/cmd/timoni/mod_vendor_crd.go
@@ -59,6 +59,8 @@ type vendorCrdFlags struct {
 
 var vendorCrdArgs vendorCrdFlags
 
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

PR #589, opened 2026-08-01
fix(cli): Preserve command cancellation

reneleonhardt:
## What

Propagate `SIGINT`, `SIGTERM`, and Cobra caller cancellation through command timeouts, Kubernetes and registry operations, completions, and Git metadata collection. Restore default signal handling after cancellation.

## Why

The logger hook and command timeouts replaced caller contexts with background contexts, delaying cancellation until local timeouts or abrupt process termination. Continued signal interception could also prevent a second signal from terminating a stuck command.

## Reproduction

```shell
nc -l 5000 >/dev/null &
server=$!
mod pull oci://127.0.0.1:5000/example -v 1.0.0 -o /tmp/module --creds : --registry-insecure --timeout 1h &
pid=$!
sleep 1
kill -TERM "$pid"
wait "$pid"
kill "$server"
# main: the process is terminated directly
# here: stderr contains 'context canceled' after deferred cleanup
```

## Verification

```shell
go test ./cmd/timoni -run 'Test(RootCommandPreservesContextCancellation|RestoreSignalHandlingAfterCancellation)' -count=1
go test ./internal/oci -run TestAppendGitMetadata -count=1
GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c ./cmd/timoni -o /tmp/timoni-command-context.test.exe
```

### DIFF
--- a/cmd/timoni/main.go
+++ b/cmd/timoni/main.go
@@ -17,9 +17,12 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"os"
+	"os/signal"
 	"path"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/fatih/color"
@@ -96,7 +99,10 @@ func init() {
 
 func main() {
 	setCacheDir()
-	if err := rootCmd.Execute(); err != nil {
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer cancel()
+	go restoreSignalHandling(ctx, cancel)
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
 		// Ensure a logger is initialized even if the rootCmd
 		// failed before running its hooks.
 		if cliLogger.IsZero() {
@@ -108,6 +114,12 @@ func main() {
 		cliLogger.Error(nil, err.Error())
 		os.Exit(1)
 	}
+}
+
+// restoreSignalHandling restores default signal behavior after cancellation.
+func restoreSignalHandling(ctx context.Context, stop context.CancelFunc) {
+	<-ctx.Done()
+	stop()
 }
 
 func setCacheDir() {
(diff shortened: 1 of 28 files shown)


cmd/timoni/main.go
```diff
@@ -96,7 +99,9 @@ func init() {
 
 func main() {
 	setCacheDir()
-	if err := rootCmd.Execute(); err != nil {
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer cancel()
```
stefanprodan, line 103: `signal.NotifyContext` continues intercepting SIGINT and SIGTERM until `cancel()` runs, which is deferred until `ExecuteContext` returns.

If execution is in a phase that does not consume the context e.g. local artifact packaging, then subsequent Ctrl-C signals are swallowed too, leaving SIGKILL as the only way to stop the process.

cmd/timoni/main_test.go
```diff
@@ -99,6 +100,19 @@ func executeCommand(cmd string) (string, error) {
 	return executeCommandWithIn(cmd, nil)
 }
 
+func TestRootCommandPreservesContext(t *testing.T) {
```
stefanprodan: Isn't this a duplicate of `TestRootCommandPreservesContextCancellation`?

cmd/timoni/main.go
reneleonhardt: Fixed in `dc20d38`. After the first cancellation, `restoreSignalHandling` calls the stop function so subsequent signals use the default OS behavior. `TestRestoreSignalHandlingAfterCancellation` covers the restoration.

cmd/timoni/main_test.go
reneleonhardt: Yes. Removed the duplicate and replaced it with the focused signal-restoration test.

stefanprodan (approved):
LGTM

PR #590, opened 2026-08-01
fix(oci): Report image build cleanup errors

reneleonhardt:
## What

Return temporary layer cleanup errors from local OCI builds and registry pushes without replacing an earlier failure.

## Why

All `ImageBuild.Close` results were discarded, so failed cleanup could be reported as success and leave temporary layer files behind.

## Reproduction

```shell
git grep 'defer build.Close()' -- cmd/timoni internal/oci
# main: four call paths discard cleanup errors
# here: no matches; each call path returns cleanup errors
```

## Verification

```shell
go test ./internal/oci -run 'Test(BuildImages|ImageBuildCloseWithError)' -count=1
go test ./cmd/timoni -run 'Test_(BuildArtifact|BuildMod|PushArtifact|PushMod)$' -count=1
GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/oci -o /tmp/timoni-oci-cleanup.test.exe
```

### DIFF
--- a/cmd/timoni/artifact_build.go
+++ b/cmd/timoni/artifact_build.go
@@ -65,7 +65,7 @@ func init() {
 }
 
 // buildArtifactCmdRun validates, builds, and writes a local artifact.
-func buildArtifactCmdRun(cmd *cobra.Command, _ []string) error {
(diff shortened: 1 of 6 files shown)


stefanprodan (approved):
LGTM

PR #612, opened 2026-08-09
feat(oci): Push pre-built module OCI archives

### LINKED ISSUE #611: Allow timoni mod push to push pre-built OCI tarballs
timoni mod build can produce a self-contained OCI .tar archive for offline builds. However, timoni mod push currently expects a local module directory rather than accepting the pre-built OCI tarball directly.

This means that after building an offline artifact, users need to push from the original module directory instead of being able to push the artifact they have already built.

Allow timoni mod push (or an appropriate command) to accept a pre-built OCI .tar archive and push it directly to an OCI registry.

reneleonhardt:
## What

Allow `timoni mod push` to push the OCI archive or layout produced by `timoni mod build` directly, without re-packaging the source module directory. The pushed digest matches the digest printed by `mod build`.

Closes #611.

## Why

`mod build` produces a self-contained OCI `.tar` for offline builds, but `timoni mod push` required the original module directory. Users who built the artifact elsewhere (sandbox, another machine) could not push what they built without extra tools like crane or skopeo.

## Reproduction

```shell
timoni mod build ./modules/my-app -v 1.0.0 -o ./my-app-1.0.0.oci.tar
timoni mod push ./my-app-1.0.0.oci.tar oci://ghcr.io/org/modules/app -v 1.0.0
# main: timoni mod push rejects the .tar and requires the module source directory.
# here: the pre-built archive is pushed as-is.
```

## Verification

```shell
go test ./cmd/timoni/ ./internal/oci/ -count=1
```

### DIFF
--- a/cmd/timoni/mod_push.go
+++ b/cmd/timoni/mod_push.go
@@ -38,7 +38,10 @@ var pushModCmd = &cobra.Command{
 	Args:  cobra.MaximumNArgs(2),
 	Short: "Push a module to a container registry",
 	Long: `The push command packages the module as an OCI artifact and pushes it to the
-container registry using the version as the image tag.`,
+container registry using the version as the image tag.
+
+The push command can also push a module OCI archive or layout produced by
+'timoni mod build' without accessing the original module directory.`,
 	Example: `  # Push a module to Docker Hub using the credentials from '~/.docker/config.json'
   echo $DOCKER_PAT | docker login --username timoni --password-stdin
   timoni mod push ./path/to/module oci://docker.io/org/app-module -v 1.0.0
@@ -78,6 +81,9 @@ container registry using the version as the image tag.`,
   timoni mod push ./path/to/module oci://ghcr.io/org/modules/app \
 	--version=1.0.0 \
 	--sign=cosign
+
+  # Push a pre-built module OCI archive produced by 'timoni mod build'
+  timoni mod push ./module-1.0.0.oci.tar oci://docker.io/org/app-module -v 1.0.0
 `,
 	RunE: pushModCmdRun,
 }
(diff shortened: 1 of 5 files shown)


internal/oci/push_module.go
```diff
@@ -47,3 +53,141 @@ func PushModule(ociURL, contentPath string, ignorePaths []string, annotations ma
...
+		if err != nil {
+			return nil, cleanup, fmt.Errorf("opening OCI archive failed: %w", err)
+		}
+		defer archive.Close()
+
+		// Archives are plain tar streams of an OCI image layout.
+		if err := tar.Untar(archive, tmpDir,
+			tar.WithSkipGzip(),
+			tar.WithSkipSymlinks(),
+			tar.WithMaxUntarSize(tar.UnlimitedUntarSize),
```
stefanprodan: Let's not repeat this mistake again, we just fixed this last week, please set here the default limit.

internal/oci/push_module.go
```diff
@@ -47,3 +53,141 @@ func PushModule(ociURL, contentPath string, ignorePaths []string, annotations ma
...
+
+	digestURL := ref.Context().Digest(digest.String()).String()
+	return fmt.Sprintf("%s%s", apiv1.ArtifactPrefix, digestURL), nil
+}
+
+// IsOCILayout reports whether path is an OCI image layout directory, that is
+// a directory written by `mod build --format=oci-layout` rather than a module
+// source directory. It is cheap to call and reads no blobs.
+func IsOCILayout(path string) bool {
+	for _, name := range []string{"oci-layout", "index.json"} {
```
stefanprodan: This feels very fragile, a module source can contain these 2. Let's drop support for layout dirs and only do OCI archives.

cmd/timoni/mod_push.go
```diff
@@ -78,6 +81,9 @@ container registry using the version as the image tag.`,
   timoni mod push ./path/to/module oci://ghcr.io/org/modules/app \
 	--version=1.0.0 \
 	--sign=cosign
+
+  # Push a pre-built module OCI archive produced by 'timoni mod build'
+  timoni mod push ./module-1.0.0.oci.tar oci://docker.io/org/app-module -v 1.0.0
```
stefanprodan: I think we should extract the version from the archive manifest instead of asking users to specify it again, so this command would become the equivalent of `helm push mychart-0.1.0.tgz oci://localhost:5000/helm-charts`.

internal/oci/push_module.go
```diff
@@ -47,3 +53,141 @@ func PushModule(ociURL, contentPath string, ignorePaths []string, annotations ma
...
+	if manifest.Config.MediaType != apiv1.ConfigMediaType {
+		_ = cleanup()
+		return nil, gcrv1.Hash{}, nil, fmt.Errorf("unsupported artifact type '%s', must be '%s'",
+			manifest.Config.MediaType, apiv1.ConfigMediaType)
+	}
+	if builtVersion, ok := manifest.Annotations[apiv1.VersionAnnotation]; ok && builtVersion != version {
+		_ = cleanup()
+		return nil, gcrv1.Hash{}, nil, fmt.Errorf("version mismatch: archive was built with version %s, cannot push as %s",
+			builtVersion, version)
+	}
```
stefanprodan, line 134: This will also match generic artifacts produced with `timoni artifact build`. I would do:

```go
  func validateModuleManifest(manifest *gcrv1.Manifest, version string) error {
    if manifest.MediaType != types.OCIManifestSchema1 {
      return fmt.Errorf("unsupported manifest media type %q", manifest.MediaType)
    }

    if manifest.Config.MediaType != apiv1.ConfigMediaType {
      return fmt.Errorf("unsupported config media type %q", manifest.Config.MediaType)
    }

    if got, ok := manifest.Annotations[apiv1.VersionAnnotation]; !ok {
      return fmt.Errorf("module version annotation is missing")
    } else if got != version {
      return fmt.Errorf(
        "version mismatch: archive was built with version %s, cannot push as %s",
        got, version,
      )
    }

    if len(manifest.Layers) != 2 {
      return fmt.Errorf("invalid module artifact: expected 2 layers, got %d", len(manifest.Layers))
    }

    expected := []string{
      apiv1.TimoniModVendorContentType,
      apiv1.TimoniModContentType,
    }

    for i, layer := range manifest.Layers {
      if layer.MediaType != apiv1.ContentMediaType {
        return fmt.Errorf("invalid module layer %d media type %q", i, layer.MediaType)
      }

      if got := layer.Annotations[apiv1.ContentTypeAnnotation]; got != expected[i] {
        return fmt.Errorf(
          "invalid module layer %d content type %q, expected %q",
          i, got, expected[i],
        )
      }
    }

    return nil
  }
```

reneleonhardt:
All four suggestions applied:

1. **Default tar limit** — `WithMaxUntarSize(UnlimitedUntarSize)` removed; extraction uses the fluxcd default size cap.
2. **Drop layout dirs** — `IsOCILayout` and the layout-directory branch removed; `mod push` archive input is an OCI archive file only; `Test_PushMod_LayoutDir` and `TestIsOCILayout` deleted.
3. **Version from manifest** — new `ModuleVersion` reads `org.opencontainers.image.version` from the archive; `-v` is optional for archives (helm-push style); `Test_PushMod_Archive` now pushes without `-v`.
4. **validateModuleManifest** — adopted verbatim, replacing the loose guards; covered by unit tests including the generic-artifact rejection.

Additional codemap review-loop work beyond the four suggestions:
- Command restructure: `-v` validation moved to the directory branch; archive branch resolves version from the manifest before building the tagged URL.
- Tests reworked to the new shape: unit tests for `ModuleVersion`, `imageFromLocal` (rejects dirs, corrupt archives), and `validateModuleManifest` (media types, version, layer count, layer content types).
- `errcheck` cleanups on deferred `cleanup()`/`archive.Close()`.
- Docs: `docs/cue/module/publishing.md` and the `mod push` help now show archive push without `-v`.
- Full `go test ./...` green; `validateModuleManifest` 94% unit-covered; `git diff --check` clean.

New commit for reviewability, use `Squash and merge` when finished.

stefanprodan (approved):
LGTM

Thanks @reneleonhardt 🏅

PR #618, opened 2026-08-10
fix(cli): Propagate bundle render output errors

reneleonhardt:
## What

Make a successful `timoni bundle build` exit guarantee that the complete in-memory render was accepted by stdout. Propagate the final writer error, and report `io.ErrShortWrite` when the writer accepts fewer bytes without returning an error. The command, flags, rendered bytes, ordering, concurrency, and existing `--output-dir` behavior stay unchanged; the `--output-dir` path already returned file write errors and is untouched.

## Why

Bundle rendering already has an excellent automation shape: Timoni builds every instance in memory, preserves bundle order, and performs one final stdout write only after the whole render succeeds. The final `Write` result is currently discarded, so a full, closed, or otherwise failing output sink can receive incomplete data while Timoni reports success. This fix turns the exit status into a dependable completeness signal for artifact pipelines and other render-only consumers, without adding a destination directory, staging tree, or apply step.

## Reproduction

The command before and after this change is identical.

```shell
# Linux: a failing output sink (writes fail with ENOSPC).
timoni bundle build -f bundle.cue >/dev/full
echo $?

# main: discards the write error, leaves the stdout stream empty, still exits 0.
# this PR: returns the write error, logs it to stderr, exits 1.
```

## What the reviewer must decide

1. Whether failing the command on a short write with a nil error (not only on a returned writer error) is the right completeness contract for a render-only stdout command.
2. Whether keeping the sibling single-instance `timoni build` on its existing behavior (it already propagates the write error but does not check short writes) is acceptable, or whether the two render commands should share the same check.

## Verification

```shell
go test ./cmd/timoni/ -run 'Test_BundleBuild' -count=1
```

The focused suite covers a sentinel writer failure, a nil-error short write, an ordinary render failure with zero output, existing stdout bytes and ordering, and unchanged directory output.

### DIFF
--- a/cmd/timoni/bundle_build.go
+++ b/cmd/timoni/bundle_build.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 	"os"
 	"path/filepath"
@@ -227,7 +228,14 @@ func runBundleBuildCmd(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
-	cmd.OutOrStdout().Write([]byte(sb.String()))
+	data := []byte(sb.String())
+	n, err := cmd.OutOrStdout().Write(data)
+	if err != nil {
+		return err
+	}
+	if n < len(data) {
+		return io.ErrShortWrite
+	}
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

Thanks @reneleonhardt

PR #622, opened 2026-08-12
feat(reconciler): Introduce pending state for in-flight upgrades

reneleonhardt:
## What

Make `timoni apply` and `timoni bundle apply` stop immediately after a failed apply stage. Previously both `ApplyInstance` implementations evaluated the inventory update and the stale-object prune before calling `kerrors.NewAggregate`, so both ran even when applying a set or waiting for readiness had failed, and a new instance stored its intended inventory before its first apply. The post-apply flow is now sequential and failure-gated: apply, record a durable transitional receipt with the exact stale identities, prune, finalize the inventory, clear the receipt.

## Why

A failed apply could leave the stored release inventory describing a render that was never applied, and the garbage collector could delete stale objects while the replacement render was incomplete. After the change, a failed apply leaves the predecessor inventory and the stale objects untouched, pruning starts only after the transitional state is durable, and the final inventory replaces the transitional state only after every stale object is proven absent. A later run resumes pruning from the recorded receipt instead of re-discovering stale objects.

## Reproduction

The behavior is visible without a cluster in a unit test: make the apply stage return a sentinel error and record calls to the inventory update and prune stages. Before the fix both later stages run; after the fix neither does, and a new instance stores no inventory.

```shell
go test ./internal/reconciler/ -count=1
```

## What the reviewer must decide

Whether pruning before the final inventory update, with resume-from-receipt semantics, is the right completeness contract for direct apply, or whether the previous order (inventory before prune) should be kept for existing instances.

## Verification

```shell
make test
```

The reconciler suite covers apply, readiness, receipt-persistence, prune, and inventory-persistence failures in both the normal and interactive paths, exact stage order, new-instance and update-with-stale flows, and crash restart with exact stale reconciliation.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -77,6 +77,11 @@ Timoni's garbage collector keeps track of the applied resources
 and prunes the Kubernetes objects that were previously applied
 but are missing from the current revision.
 
+On upgrade, Timoni records the new revision as pending in the instance
+storage before touching the cluster, and marks it as applied only after
+the reconciliation has succeeded. If an upgrade is interrupted, the next
+run resumes from the recorded state, and `timoni delete` cleans up both revisions.
+
 After an installation or upgrade, Timoni waits for the
 applied resources to be fully reconciled by checking the ready status
 of deployments, jobs, services, ingresses, and Kubernetes custom resources.
--- a/cmd/timoni/bundle_delete.go
+++ b/cmd/timoni/bundle_delete.go
@@ -168,8 +168,8 @@ func deleteBundleInstance(ctx context.Context, instance *apiv1.BundleInstance, w
 		return err
 	}
 
-	iManager := runtime.InstanceManager{Instance: *inst}
-	objects, err := iManager.ListObjects()
+	// Cover the pending revision of an unfinished upgrade too.
+	objects, err := iStorage.ListAllObjects(ctx, instance.Name, instance.Namespace)
 	if err != nil {
 		return err
 	}
--- a/cmd/timoni/delete.go
+++ b/cmd/timoni/delete.go
@@ -100,8 +100,8 @@ func runDeleteCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	iManager := runtime.InstanceManager{Instance: *inst}
-	objects, err := iManager.ListObjects()
+	// Cover the pending revision of an unfinished upgrade too.
+	objects, err := iStorage.ListAllObjects(ctx, deleteArgs.name, *kubeconfigArgs.Namespace)
 	if err != nil {
 		return err
 	}
--- a/internal/reconciler/interactive.go
+++ b/internal/reconciler/interactive.go
@@ -25,7 +25,6 @@ import (
 	"cuelang.org/go/cue"
 	"github.com/fluxcd/pkg/ssa"
 	"github.com/go-logr/logr"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/stefanprodan/timoni/internal/dyff"
 	"github.com/stefanprodan/timoni/internal/engine"
@@ -44,6 +43,15 @@ func NewInteractiveReconciler(log logr.Logger, copts *CommonOptions, iopts *Inte
 
 	if iopts.ProgressStart != nil {
 		reconciler.progressStartFn = iopts.ProgressStart
+	}
+
+	// Re-bind the post-apply stage seams so that the interactive wait
+	// implementations are used for readiness and termination.
+	reconciler.applySetsFn = func(ctx context.Context, log logr.Logger) error {
+		return reconciler.ApplyAllSets(ctx, log, reconciler.Wait)
+	}
+	reconciler.pruneStaleFn = func(ctx context.Context, log logr.Logger) error {
+		return reconciler.PostApplyPruneStaleObjects(ctx, log, reconciler.WaitForTermination)
 	}
 
 	return reconciler
@@ -82,13 +90,15 @@ func (r *InteractiveReconciler) ApplyInstance(ctx context.Context, log logr.Logg
 	} else {
 		log.Info(fmt.Sprintf("upgrading %s in namespace %s",
 			logger.ColorizeSubject(r.Name()), logger.ColorizeSubject(r.Namespace())))
+
+		if !sameInventory(r.instanceManager.Instance.Inventory, r.predecessorInventory) {
+			if err := r.savePendingFn(ctx); err != nil {
+				return fmt.Errorf("recording pending revision failed: %w", err)
+			}
+		}
 	}
 
-	return kerrors.NewAggregate([]error{
-		r.ApplyAllSets(ctx, log, r.Wait),
-		r.PostApplyUpdateInventory(ctx, builder, buildResult),
-		r.PostApplyPruneStaleObjects(ctx, log, r.WaitForTermination),
-	})
+	return r.applyInstanceStages(ctx, log, builder, buildResult)
 }
 
 func (r *InteractiveReconciler) DryRunDiff(ctx context.Context, namespaceExists bool) error {
--- a/internal/reconciler/reconciler.go
+++ b/internal/reconciler/reconciler.go
@@ -18,7 +18,9 @@ package reconciler
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sort"
 	"time"
 
 	"cuelang.org/go/cue"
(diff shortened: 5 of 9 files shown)


docs/concepts.md
```diff
@@ -77,6 +77,15 @@ Timoni's garbage collector keeps track of the applied resources
...
 but are missing from the current revision.
 
+The post-apply flow is failure-gated. After a successful apply and readiness
+wait, Timoni records a durable transitional state with the exact stale-object
+identities, prunes the stale objects, and only then replaces the stored
+inventory with the applied revision. If applying, waiting, recording or
+pruning fails, the command stops immediately: the stored inventory still
+describes the previous applied revision, stale objects are not deleted, and a
+later run resumes pruning from the recorded transitional state instead of
+re-discovering stale objects.
```
stefanprodan: This is Al speak. Better delete all of it, an end-user doesn't care about the internals of Timoni's GC.

stefanprodan (commented):
There are many issues with this PR, see https://github.com/stefanprodan/timoni/pull/622#issuecomment-5269541851

stefanprodan:
@reneleonhardt reading the code again, I found another major issue. In Flux, we prune after apply and before waiting, this is a must, as renaming a workload holding a hostPort, lease, or PV will deadlock. 

Here is the updated plan:


### Storage

Keep the single `timoni.<instance>` Secret with two data keys:

- `instance`: the applied instance record, exactly as today.
- `pending`: the instance record of a revision that is being applied but is not finalized, in the same encoding as `instance` (module reference, values, inventory, lastTransitionTime). No new type or marshaling; tools can surface an in-flight upgrade from it.

One field manager writes the Secret and always server-side-applies the full intended `data` state, so every write is atomic:

- Record: apply `{instance: <stored bytes, unchanged>, pending: <new record>}`.
- Finalize: apply `{instance: <new record>}`, which promotes the pending revision and drops the `pending` key in the same request.

The pending record stores no stale identities and no predecessor copy: the `instance` key is the predecessor, and stale objects are always recomputed (see below), so a recorded list can never go out of sync with the current render.

### Install (no stored instance)

Create the namespace, store the intended inventory, then apply and wait. No pending record, no prune. If the apply fails, the stored inventory describes the intended render, so `timoni delete` can clean up whatever subset was created (deleting missing objects is a no-op).

### Upgrade (stored instance exists)

1. Load the stored instance (predecessor) and the pending record if one exists.
2. Compute `stale = (predecessor.inventory ∪ pending.inventory) \ currentDesired`. Because the current render is always subtracted, an object in the desired state can never be pruned, and objects created by an earlier unfinished run are covered via its recorded inventory.
3. If the current desired inventory differs from the predecessor's, write the pending record before applying anything. If they are equal, skip it: a failed apply cannot untrack anything and there is nothing to prune.
4. Apply the sets with their per-set readiness waits. An apply error stops the run: nothing is pruned, the predecessor inventory and the receipt are untouched.
5. Prune the stale objects and wait for termination whenever every apply succeeded, even if a readiness wait failed. Stale objects are not part of the desired render, so removing them cannot break unapplied sets, and it unblocks replacements whose readiness is held up by the object they replace (a renamed workload keeping a hostPort, lease, or volume). Requires apply and wait errors to be distinguishable; the wait callbacks wrap their errors in a typed error for this.
6. Finalize with the single atomic write above only when apply, readiness, and prune all succeeded. After a readiness failure with a successful prune, the run reports the wait error and keeps the pending record; the retry re-applies, finds nothing new to prune, and finalizes once readiness is met.

Prune is gated on apply success, not on readiness: pruning without knowing what was applied is never safe, while pruning before readiness only exposes removed identities, which re-applying the previous render recreates. In-place updates are overwritten by the apply itself under any ordering.

### Delete

`timoni delete` deletes the union of the stored and pending inventories (when a pending record exists), then removes the Secret. This covers objects created by an unfinished upgrade.

### Failure behavior

Any crash or error leaves one of two durable states: predecessor inventory alone, or predecessor plus a pending record naming the revision in flight. The next run, with any render, recomputes stale from that state, prunes whatever is actually stale, and finalizes. Pruning already-deleted or never-created objects is tolerated, so every intermediate state converges. Older Timoni versions ignore the extra data key.

stefanprodan:
Fable findings:

 ## Blocking

  1. The interactive path never records the pending revision. The upgrade branch that calls savePendingFn exists only in the base Reconciler.ApplyInstance
  (internal/reconciler/reconciler.go:179). InteractiveReconciler.ApplyInstance (internal/reconciler/interactive.go:60) shadows it: it handles the install case
  itself, logs "upgrading" for the else case, and jumps straight to applyInstanceStages without ever checking sameInventory or saving the pending record. Since
  both timoni apply (cmd/timoni/apply.go:250) and timoni bundle apply (cmd/timoni/bundle_apply.go:355) construct NewInteractiveReconciler, no real CLI run ever
  writes the pending key. Step 3 of your plan is unreachable outside the unit tests, and with it the crash-resume and delete-covers-unfinished-upgrade
  guarantees. The test gap mirrors the bug: TestInteractiveApplyInstallStoresIntendedInventoryFirst covers install only; there is no interactive upgrade test
  asserting the pending record. The fix is to hoist the install/pending pre-apply logic into a shared helper (or into applyInstanceStages itself) that both
  ApplyInstance implementations go through, plus an interactive upgrade test.

  2. SavePending strips the bundle-owner label from the storage Secret. StorageManager.Apply merges instance.Labels (including bundle.timoni.sh/name for
  bundle-owned instances) into the Secret via SSA under the Timoni field manager. SavePending (internal/runtime/storage.go:204) applies a Secret built from
  newSecret, which carries only the three base labels, with the same field manager and ForceOwnership, so the server removes the bundle label. I verified this
  with a scratch test against the PR head: after Apply with a bundle label followed by SavePending, the label is gone (the controller-runtime v0.24 fake client
  reproduces the SSA removal, so this is testable without envtest). Consequences for the whole pending window, which after a crash is unbounded: the instance
  vanishes from bundle-scoped listing (bundle status, bundle delete walk instances by that label), and the ownership check in Init
  (storedInstance.Labels[apiv1.BundleNameLabelKey]) no longer sees a bundle owner, so a standalone timoni apply can take over a bundle-owned instance without
  --overwrite-ownership. Fix: carry the existing Secret's labels forward in SavePending (copy existing.Labels over the base labels, the same way Apply merges
  instance.Labels).

  Note these compound: fixing 1 without 2 activates the label loss on every real upgrade.

##  Minor

  - In applyInstanceStages, when a readiness failure is followed by a prune failure, only the prune error is returned and the readiness error is swallowed. The
  old aggregate reported both; errors.Join(err, pruneErr) would preserve that.
  - The added docs/concepts.md paragraph contains design rationale ("because stale objects are not part of the desired render and removing them unblocks
  replacements held up by...") which per your docs convention belongs in the PR, not the docs. Trimming it to behavior statements would also make it less
  internal-sounding ("pending record" as a term needs no justification clauses on the concepts page).
  - sameInventory doesn't use its receiver; it can be a plain function next to mergeObjects.
  - storeInventory in the reconciler test ignores the error returned by storageManager.Apply.

reneleonhardt:
Thanks for the review — both blockers were already addressed in `cd67db4`:

- The interactive path now records the pending revision. `InteractiveReconciler.ApplyInstance` checks `sameInventory` and calls `savePendingFn` in the upgrade branch, matching the base path, with a regression test (`TestInteractiveApplyUpgradeRecordsPending`).
- `SavePending` now carries the existing Secret labels forward, so the bundle-owner label survives the pending write (`TestSavePendingKeepsSecretLabels`).

And thanks for the minors — all fixed in `9ea0204`:

- Readiness and prune failures are joined, so a prune error no longer swallows the wait error.
- `sameInventory` is a plain function instead of a method with an unused receiver.
- Trimmed the design rationale from the concepts page; it now states behavior only.
- The test helper asserts the storage error instead of ignoring it.

docs/concepts.md
```diff
@@ -77,6 +77,17 @@ Timoni's garbage collector keeps track of the applied resources
...
+On upgrade, Timoni stores the intended revision as a pending record in the
+same Secret before touching the cluster. Stale objects are recomputed from
+the stored and pending inventories minus the desired render, so an object in
+the desired state is never pruned and objects created by an unfinished run
+are still covered. Pruning runs whenever the apply succeeded, even if a
+readiness wait failed. The pending record becomes the applied inventory only
+when apply, readiness and pruning all succeeded; a crash at any point leaves
+the stored or pending record intact, and the next run prunes whatever is
+still stale and finalizes. `timoni delete` removes the objects of the stored
+and pending inventories together.
```
stefanprodan:
```suggestion
On upgrade, Timoni records the new revision as pending in the instance
storage before touching the cluster, and marks it as applied only after
the reconciliation has succeeded. If an upgrade is interrupted, the next
run resumes from the recorded state, and `timoni delete` cleans up both revisions.
```

stefanprodan:
Please rebase and squash all commits, after getting my suggestion in.

reneleonhardt:
Done. But why such a heavy kind setup instead of kubesolo? 😞
https://github.com/marketplace/actions/setup-kubesolo

stefanprodan:
All my projects use upstream tooling, kind is how Kubernetes team does testing.

stefanprodan (approved):
LGTM

Thanks @reneleonhardt for this great improvement 🥇

PR #623, opened 2026-08-12
fix(cli): Retain inventory until delete finalizes

reneleonhardt:
## What

`timoni delete` and `timoni bundle delete` now keep the release-state record (`Secret/timoni.<instance>`) until every owned object is confirmed absent, instead of dropping it right after the delete requests are issued.

- `StorageManager.SetDeleting` merges a `status.timoni.sh/deleting` receipt annotation into the storage Secret, leaving the instance data and exact inventory intact.
- The record is removed only after `WaitForTermination` succeeds, or when nothing was deleted at all.
- With `--wait=false` the inventory storage is deleted without waiting for finalizers

## Why

Deleting the inventory before termination was verified broke recovery: on a termination timeout, client cancellation, or a stuck finalizer, live objects could remain after the inventory needed to resume their cleanup was already gone. A restarted delete then failed with "instance storage not found" and could no longer reconstruct the exact object list from release state.

A restarted delete reconciles the exact retained inventory and does not resend delete requests for already-absent objects, since `ssa.Delete` treats `NotFound` as `Deleted` without issuing a request.

## Verification

```shell
make install-envtest
KUBEBUILDER_ASSETS="$(bin/setup-envtest use -i --bin-dir=bin -p path)" \
  go test ./cmd/timoni -run 'TestDelete|Test_BundleDelete' -count=1
```

Covers stuck-finalizer timeout with resumable inventory, resume after the finalizer clears, `--wait=false` retention, state removal when nothing was deleted, and foreign same-label objects staying untouched, for both `timoni delete` and `timoni bundle delete`. The full test suite (`go test ./cmd/timoni ./internal/... ./api/...`) passes.

### DIFF
--- a/cmd/timoni/bundle_delete.go
+++ b/cmd/timoni/bundle_delete.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"sort"
 
 	"cuelang.org/go/cue/cuecontext"
@@ -39,7 +38,12 @@ var bundleDelCmd = &cobra.Command{
 	Short:   "Delete all instances from a bundle",
 	Args:    cobra.MaximumNArgs(1),
 	Long: `The bundle delete command uninstalls the instances and
-deletes all their Kubernetes resources from the cluster.'.
+deletes all their Kubernetes resources from the cluster.
+
+The release-state record of each instance is removed only after its owned
+objects are confirmed absent. If the termination cannot be verified
+(--wait=false, a timeout, or an error), the record is retained as a
+deletion-in-progress receipt and a subsequent delete finalizes the removal.
 `,
 	Example: `  # Uninstall all instances in a bundle
   timoni bundle delete -f bundle.cue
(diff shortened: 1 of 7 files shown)


cmd/timoni/bundle_delete.go
```diff
@@ -39,7 +38,12 @@ var bundleDelCmd = &cobra.Command{
 	Short:   "Delete all instances from a bundle",
 	Args:    cobra.MaximumNArgs(1),
 	Long: `The bundle delete command uninstalls the instances and
-deletes all their Kubernetes resources from the cluster.'.
+deletes all their Kubernetes resources from the cluster.
+
+The release-state record of each instance is removed only after its owned
+objects are confirmed absent. If the termination cannot be verified
+(--wait=false, a timeout, or an error), the record is retained as a
+deletion-in-progress receipt and a subsequent delete finalizes the removal.
```
stefanprodan: When `--wait=false` Timoni should never stall or leave the storage in-cluster, setting this flag must match kubectl behavior (no waiting on finalizers). 

Also please rewrite the text here and make it for humans, reads like AI Slop with jargon that even I can not grok.

reneleonhardt: Fixed, please review again.

`--wait=false` now behaves like kubectl: it sends the delete requests, removes the instance record right away, and never waits on finalizers or leaves storage behind.

In addition:
- Rewrote the `delete` and `bundle delete` help text in plain language.
- Reworded the related docs and code comments the same way.
- Added tests for both paths: `--wait=false` dropping the record without waiting, and `--wait` keeping the record on timeout so the delete can be retried.

stefanprodan (approved):
LGTM

Thanks @reneleonhardt

reneleonhardt:
You're welcome! Believe me, it's not easy to steer those agents, I let them proof-read and review in dozens of loops 😅
Do you need a better PR description?

stefanprodan:
I amended the PR to say what `--wait=false` does after the last commit.

What I really want to avoid is having user-facing docs and help text being polluted with AI slop. Agents tend to generate dense docs with jargon that they develop while writing the code, which in almost all cases is not what an end-user needs.

reneleonhardt:
I couldn't agree more, I saw more than enough already, can't read that myself and ask for a summary 😄

PR #654, opened 2026-09-09
feat: Update bundles from local sources

reneleonhardt:
## What

Add two independent local-source modes to `timoni bundle update`, plus an opt-in atomic update-and-render workflow:

1. `--local-index` maps exact OCI module identities and semantic versions to extracted `file://` sources with pinned SHA-256 source-tree digests. The updater verifies the selected source before planning, rewrites the source, version, and digest together, and passes the digest through bundle apply so local fetches are verified again before rendering.
2. Repeatable `--oci oci://repository=path` mappings select local OCI archives or image layouts. Timoni verifies the manifest version and artifact digest before planning and fetching, then rewrites the selected source to a canonical `file://` path.
3. `bundle build --update` stages the selected changes in memory, vets and builds those staged inputs, buffers rendered stdout, and writes bundle files only after validation succeeds. Ordinary `bundle build` and `bundle update` behavior remains unchanged; the atomic mode rejects `-f -` and `--output-dir`.

The local-source modes work independently and may be combined intentionally. A bundle may mix remote OCI references, local OCI artifacts, and ordinary mutable local modules. With `--local-index`, a matching OCI reference may switch to its indexed local source; an unindexed OCI reference keeps the registry fallback.

## Why

Local modules were skipped because a filesystem path does not prove which published module and version it represents. The explicit index supplies that authority while deterministic hashing detects stale or tampered sources. Regular files and symlink target strings are hashed; other special files are rejected. Local OCI mappings supply the missing repository identity explicitly because an archive or image layout contains its version and artifact digest, but not the registry repository name.

### 1. Update an extracted local module with `--local-index`

Before this change, a `file://` reference was skipped because it did not carry a published OCI identity:

```cue
module: url: "file://../modules/app"
```

```console
$ timoni bundle update -f bundle.cue --dry-run
file:// instance -> skipped: module is not an OCI artifact
```

The workaround was to replace the checkout with the desired extracted archive:

```console
$ tar -xf app.tar -C ../modules/app
$ timoni mod vet ../modules/app --name app --namespace app
$ timoni bundle vet --runtime-from-env -f bundle.cue
$ timoni bundle build --runtime-from-env --mask-secrets -f bundle.cue > render.yaml
$ less render.yaml
```

Use an explicit index beside the extracted module sources; this branch intentionally does not generate it.

```cue
modules: {
	"oci://registry.io/app": versions: {
		"1.1.0": {
			source: "file://../modules/app-1.1.0"
			digest: "sha256:<64-hex-digest>"
		}
	}
}
```

Replace `<64-hex-digest>` with the source-tree digest, then run the update first as a dry run and apply it:

```console
$ timoni bundle update --local-index module-index.cue -f bundle.cue --dry-run
# reports the source, version, and digest transition without writing
$ timoni bundle update --local-index module-index.cue -f bundle.cue
```

The index supplies the OCI identity, semantic version, source path, and source-tree digest. The updater selects and rewrites all three bundle literals together, and the shared fetcher verifies the selected source again before rendering. Run the normal `bundle vet` and `bundle build` commands afterward, or use the atomic workflow below.

Relative sources resolve from the index file; filesystem paths never determine the OCI identity. A source mutation, missing source, invalid special file, or digest mismatch causes the update to fail before the bundle changes. The required invariant is:

```text
indexed version/digest -> selected source -> rendered selected module
```

### 2. Update directly from a local OCI artifact with `--oci`

Use the artifact itself as the source and map it explicitly to its repository:

```console
$ timoni bundle update --oci oci://registry.io/app=app.oci.tar -f bundle.cue --dry-run
# verifies the local artifact and reports the available update without writing
$ timoni bundle update --oci oci://registry.io/app=app.oci.tar -f bundle.cue
```

The `--oci` mapping also accepts an OCI image-layout directory and is repeatable for multiple repositories. Path-only inputs are rejected because Timoni never infers an OCI identity from a filename or argument order. The manifest version and artifact digest are checked before planning and again when the selected artifact is fetched for rendering. This mode works without `--local-index`, just as the indexed-source mode works without `--oci`; use the normal vet/build steps afterward or the atomic workflow below.

### 3. Update and render atomically with `bundle build --update`

The update-only workflow still leaves validation and rendering as separate commands:

```console
$ timoni bundle update --local-index module-index.cue -f bundle.cue
$ timoni bundle vet -f bundle.cue
$ timoni bundle build -f bundle.cue --mask-secrets
```

When the build is the desired target, the opt-in transaction combines those steps while consuming the same staged source selection:

```console
$ timoni bundle build --update --local-index module-index.cue -f bundle.cue --runtime-from-env --mask-secrets
$ timoni bundle build --update --oci oci://registry.io/app=app.oci.tar -f bundle.cue --runtime-from-env --mask-secrets
```

The command loads only explicitly authorized local sources, plans and verifies the selected versions, stages the updated CUE files in memory, runs vet and build against those staged inputs, and buffers rendered stdout. It writes the changed bundle files and exposes the render output only after every step succeeds. If planning, source verification, vet, build, render, or the final multi-file write fails, the original bundle files remain byte-for-byte unchanged. The workflow does not apply anything to a cluster or publish to a registry.

## Verification

```shell
go test ./internal/engine/... ./internal/oci -count=1
go test ./cmd/timoni -run '^(Test_BundleUpdate|Test_BundleUpdateLocalOCI|Test_BundleBuildUpdateMixedLocalSources)$' -count=1
go test ./... -count=1
```

The focused and black-box suites cover indexed version selection, digest verification, cache-hit revalidation, symlink target changes, imported CUE package files, dry-run behavior, OCI-to-local substitution, local rendering, mixed local-source updates, CLI validation, and the existing OCI path.

### DIFF
--- a/api/v1alpha1/module.go
+++ b/api/v1alpha1/module.go
@@ -20,8 +20,8 @@ package v1alpha1
 // denotes the latest stable version of a module.
 const LatestVersion = "latest"
 
-// ModuleReference contains the information necessary to locate
-// a module's OCI artifact in the registry.
+// ModuleReference contains the information needed to locate a module artifact
+// or local source.
 type ModuleReference struct {
 	// Name of the module.
 	Name string `json:"name"`
@@ -30,10 +30,11 @@ type ModuleReference struct {
 	// 'oci://<reg.host>/<org>/<repo>' or 'file://<path>'.
 	Repository string `json:"repository"`
 
-	// Version is the OCI artifact tag in strict semver format.
+	// Version is the strict semantic version of the module release, when known.
 	Version string `json:"version"`
 
-	// Digest of the OCI artifact in the format '<sha-type>:<hex>'.
+	// Digest is an OCI artifact digest or the source-tree digest of an indexed
+	// local module, in the format '<sha-type>:<hex>'.
 	Digest string `json:"digest"`
 
 	// Annotations of the OCI artifact.
--- a/cmd/timoni/bundle_apply.go
+++ b/cmd/timoni/bundle_apply.go
@@ -313,6 +313,7 @@ func fetchBundleInstanceModule(ctx context.Context, instance *apiv1.BundleInstan
 		f, err := fetcher.New(ctx, fetcher.Options{
 			Source:      instance.Module.Repository,
 			Version:     moduleVersion,
+			Digest:      instance.Module.Digest,
 			Destination: dstDir,
 			CacheDir:    rootArgs.cacheDir,
 			Creds:       creds,
--- a/cmd/timoni/bundle_build.go
+++ b/cmd/timoni/bundle_build.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 	goruntime "runtime"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
@@ -50,6 +51,8 @@ var bundleBuildCmd = &cobra.Command{
 	Long: `The bundle build command builds and prints the resulting Kubernetes resources for all instances defined in a Bundle.
 
 Custom resources are validated against the CRD schemas vendored in the modules and the CRDs rendered by any instance of the bundle. A violation fails the command and nothing is written. Use --validate=false to disable the validation.
+
+With --update, module references are updated, vetted, and built from the same in-memory inputs. Bundle files are written only after vet and build succeed. Use --local-index for extracted local modules and repeatable --oci mappings for local OCI artifacts.
 `,
 	Example: `  # Build all instances from a bundle and print the manifests to stdout
   timoni bundle build -f bundle.cue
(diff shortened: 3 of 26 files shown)


stefanprodan:
Hey @reneleonhardt I'm quite confused about what this is. If a module is local, the bundle caries no version/digest info, so why would you update it? Local modules are not versioned in a registry, they are meant to be versioned in Git. I would really prefer for you to open an issue with the feature request before a 2K LoC PR to discuss the feasibility.

reneleonhardt:
Your concern is valid. The local OCI archive workaround was:

```bash
timoni bundle build -f bundle.cue --module-source my.oci.tar --runtime-from-env --mask-secrets
```

That requires materializing `my.oci.tar` from the committed module first.

After, with `bundle.cue` pointing directly to the committed local module:

```bash
timoni bundle build -f bundle.cue --runtime-from-env --mask-secrets --vet
```

The new `--vet` flag wraps the existing validation phase and normal build in one read-only command. It reuses the same inputs, resolves runtime values once, reads stdin once, and emits no manifests if validation fails.

This does not add bundle update semantics, local indexes, synthetic versions, digests, or OCI archives. `--validate` keeps its existing CRD validation meaning, and `--vet` does not replace `mod vet`.

I implemented this, if you want, I can replace this PR or create a new one.

`feat(bundle): Validate before building manifests`

- Prod: 2 files, `+73 -49`
- Tests: 1 file, `+33 -5`
- Docs/skill: 2 files, `+12 -2`
- Total: 5 files, `+118 -56`

stefanprodan:
> The new --vet flag wraps the existing validation phase and normal build in one read-only command.

Hmm but `bundle build` does vetting of the bundle definition by design when it materializes the cue object.

PR #657, opened 2026-09-16
fix: remove trailing c-directives-end marker from timoni build YAML output

### LINKED ISSUE #656: `timoni build` adds a trailing document marker '---'
The `---` marker separates a YAML document's directives from its content. `timoni build` always appends one to the end of the output stream even though no additional documents follow.

To reproduce:

```bash
cd $(mktemp -d)
timoni mod init foo
timoni build bar foo | tail -n 5
```

Which outputs:
```
          - key: index.html
            path: index.html
          name: bar-db0499e2
        name: html
---
```

kormide:
Don't add a trailing yaml document separator `---` in the `timoni build` output. The separator indicates the start of a document, and placing one at the end is inconsistent with other k8s tooling.

Closes https://github.com/stefanprodan/timoni/issues/656.


cmd/timoni/build.go
```diff
@@ -243,7 +243,6 @@ func runBuildCmd(cmd *cobra.Command, args []string) error {
 				return fmt.Errorf("converting objects failed: %w", err)
 			}
 			sb.Write(data)
-			sb.WriteString("---\n")
```
stefanprodan, line 246: What happens if there are multiple objects?

kormide: Oh, I misunderstood and thought that the above line output all of the manifests with their separators, and that this line added a final one. I'll do another pass. Not sure why I didn't notice the loop 😅 .

kormide: I think the solution is to just move the `---` before each manifest is printed.

cmd/timoni/build_test.go
```diff
@@ -54,6 +54,41 @@ func TestBuild(t *testing.T) {
 		}
 	})
 
+	t.Run("builds module without preceding yaml document separator", func(t *testing.T) {
```
seh:
```suggestion
	t.Run("builds module without preceding YAML document separator", func(t *testing.T) {
```

cmd/timoni/build_test.go
```diff
@@ -54,6 +54,41 @@ func TestBuild(t *testing.T) {
...
+			name,
+			modPath,
+		))
+
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(output).ToNot(HavePrefix("---"))
+	})
+
+	t.Run("builds module without trailing yaml document separator", func(t *testing.T) {
+		// The yaml document separator "---" marks the beginning of a document, so it
```
seh:
```suggestion
		// The YAML document separator "---" marks the beginning of a document, so it
```

cmd/timoni/build_test.go
```diff
@@ -54,6 +54,41 @@ func TestBuild(t *testing.T) {
 		}
 	})
 
+	t.Run("builds module without preceding yaml document separator", func(t *testing.T) {
+		// It is perfectly valid to begin with a preceding "---", however it's common
```
seh: This is not about common practice, but rather about the YAML specification.

The document separator is separating a document's directives from its content, per YAML 1.2.2 [§9.1.2](https://yaml.org/spec/1.2.2/#912-document-markers). As a degenerate alternative, per [§9.1.4](https://yaml.org/spec/1.2.2/#914-explicit-documents), a `c-directives-end` ("---") can start a document with no directives.

Since starting a document ends a previous document implicitly, it looks like a document separator, but that's not really what it exists to separate.

kormide: Thanks. I revised the explanation.

